Issue #35, opened 2026-06-15
Auto-reconnect loop has no backoff and no give-up — drains battery hunting a ring that's off/charging/out of range


bluna-ai:
## Problem
`RingScanner.didDisconnectPeripheral` (`BLE/RingScanner.swift:299-315`) immediately re-issues `central.connect(peripheral)` whenever `wantConnection` is true, with **no backoff, no cap, and no awareness that the ring is deliberately off** (on the charger, removed for the day, dead battery). CoreBluetooth's connect-with-no-timeout will keep the radio armed indefinitely.

While CoreBluetooth itself coalesces pending connects efficiently, the app re-arms on every disconnect and also re-arms in `endBackgroundReadRearming` and `centralManagerDidUpdateState`. Combined with the 30s keepalive (#31), a ring that's charging overnight = the phone holding an armed BLE connect + periodic scan attempts for hours.

There's also no user-visible "ring unreachable for X / on charger" state — the UI just says "Connecting…" forever (`statusText` `:483`), which is confusing during normal charging.

## Proposed fix
- Add exponential backoff on repeated immediate disconnects (e.g. ring rejects/charging): 1s → 5s → 30s → cap, reset on a successful data frame.
- Detect "intentional off" (charging) heuristics and surface a calm state ("Ring charging / out of range — will reconnect automatically").
- Keep the standing pending connect (cheap) but stop active re-scans during the backoff.

Priority: **P2**

perezjuanj:
Done in #51: exponential reconnect backoff with a cap + reset-on-success (`ReconnectBackoff`, unit-tested) + a calm 'unreachable' state. Deferred follow-up: a watchdog for the never-connected (no didDisconnect) case.

PR #64, opened 2026-06-17
Battery freshness indicator + honest connectivity/charging copy (#57, #60)

### LINKED ISSUE #57: Battery % reads as live while minutes stale — tighten freshness indicator
## Problem
Battery % is only greyed / annotated "as of Xm ago" after the idle staleness threshold `idleStaleAfter = 360s`. So a battery value up to ~6 minutes old still renders as if it were current. On a glanceable home card this is a trust gap — the headline number looks live when it isn't.

## Proposed fix
Apply a tighter, dedicated freshness window to battery (independent of the broad live-readings stale window): grey it and show "as of Xm ago" once the last 0x10/0x87 descriptor is older than ~the keepalive interval (e.g. ~60-120s), and clear it immediately on a fresh fetch. Battery itself is 🟢 ground-truthed (byte[1]); this is purely a freshness-presentation change.

## Acceptance criteria
- Battery shows an "as of Xm ago" caption / greying within ~2 minutes of the last descriptor read.
- A fresh descriptor clears the staleness annotation.
- Low-battery coloring still works.

## File pointers
- ios/OpenRingConn/BLE/RingSession.swift:83-89 (`idleStaleAfter = 360`)
- ios/OpenRingConn/ContentView.swift:193-204 (battery render), :193-270 (battery card)
- ios/OpenRingKit/Sources/OpenRingKit/DeviceStatus.swift:31-35 (battery decode 🟢)

## Confidence
Battery 🟢. No protocol work; presentation only.

---
**Priority:** P2 · **Complexity:** low · **Recommended impl model:** haiku · **Actionable now:** yes
<sub>Filed from the APK-parity + UX/reliability triage round. Implemented together with the "honest charging copy" ticket in the battery/connectivity lane.</sub>

### LINKED ISSUE #60: Stop the reconnect status from falsely claiming "charging"; honest connectivity copy
## Problem
The maintainer asked to "show when charging." Today the status text does the opposite of honest: on repeated failed reconnects it shows "Ring unreachable or charging — will reconnect automatically", derived **purely from the reconnect attempt count** (`reconnectStalled`). No charging byte is decoded, and the code itself flags this is NOT a charging signal (ReconnectBackoff.swift:9-10). So out-of-range and on-charger are conflated, and the app asserts a state it cannot know.

The real charging-byte decode is blocked on protocol RE and is tracked by #41 (candidate 0x10/0x87 [2] state enum is 🔴 guess). Reference #41 for the true signal; this ticket is the interim honesty fix, which #41 does not cover.

## Proposed fix
- Reword to an honest connectivity state that does not assert charging, e.g. "Ring unreachable — reconnecting automatically."
- OPTIONALLY show a tentative "Charging" hint **only** from the proven battery proxy: battery % rising across recent 0x10/0x87 descriptors (battery is 🟢). Label any such hint as inferred. Never derive charging from the reconnect count.
- Leave the canonical charging byte to #41.

## Acceptance criteria
- Status no longer claims "charging" based on reconnect attempts alone.
- If a charging hint is shown, it is backed by a rising battery trend and worded as inferred.
- Out-of-range still reads as a calm, non-stuck reconnecting state (#35 behavior preserved).

## File pointers
- ios/OpenRingConn/ContentView.swift:514-520 (`statusText` — conflated "charging" copy)
- ios/OpenRingConn/BLE/RingScanner.swift:77-81 (`reconnectStalled` source)
- ios/OpenRingKit/Sources/OpenRingKit/ReconnectBackoff.swift:9-10 (explicit "NOT a charging-flag byte")
- ios/OpenRingKit/Sources/OpenRingKit/DeviceStatus.swift:20-49,31-35 (battery 🟢, for the optional rising-trend proxy)

## Confidence
Charging-from-reconnect 🔴 (remove); battery proxy 🟢; real charging byte protocol-blocked (#41).

---
**Priority:** P2 · **Complexity:** low · **Recommended impl model:** sonnet · **Actionable now:** yes · **Related:** #41
<sub>Filed from the APK-parity + UX/reliability triage round. Implemented together with the battery-freshness ticket in the battery/connectivity lane.</sub>

perezjuanj:
## Summary

- **#57 Battery freshness**: The battery % in the connection header now has its own dedicated freshness window (120 s, ~2× the night keepalive interval) tracked by a new `batteryFetchedAt` timestamp. Previously the battery used `liveReadingsStale` / `lastFrameAt`, which refreshes on every frame — including 2-s live-HR polls — so a battery reading from 5 minutes ago could silently display as "current" during a long monitoring session. `batteryStale` drives the grey-out and "as of Xm ago" label; `batteryFetchedAt` provides the accurate reference timestamp. The battery decode in `DeviceStatus` (🟢 byte[1]) is unchanged.

- **#60 Honest connectivity copy**: "Ring unreachable or charging — will reconnect automatically" asserted charging from the reconnect count alone, which `ReconnectBackoff.swift` explicitly calls out as NOT a charging signal. The new text is "Ring unreachable — reconnecting automatically". An optional honest suffix "· inferred charging" is appended only when the last session's battery % trend was strictly rising (🟢 proxy: rising % across 0x10/0x87 descriptors → ring may be charging). This inference is computed by the new pure kit function `ChargingInference.inferred(from:)`, persisted to `UserDefaults` in `RingSession.invalidate()` (so it's readable during the reconnect-backoff window when session == nil), and labeled "inferred" throughout to avoid overstating certainty from an unconfirmed protocol field.

Closes #57
Closes #60

## Kit build / test results

```
swift build: Build complete! (3.73s)
swift test:  Executed 134 tests, with 0 failures (0 unexpected)
             (includes 13 new ChargingInferenceTests)
```

## New tests added

`ios/OpenRingKit/Tests/OpenRingKitTests/ChargingInferenceTests.swift` — 13 cases covering:
- empty / single-reading edge cases (both false)
- strictly rising 2-, 3-, 4-reading windows (true)
- flat, falling, and mixed (non-all-rising) windows (false)

## On-device verification needed

- **Battery grey-out timing**: after connecting and receiving a battery reading, wait 2+ minutes without triggering a manual action — battery % should grey out and show "as of Nm ago". Reconnecting should clear it within one keepalive cycle.
- **Honest status text**: disconnect from the ring (or simulate ring-out-of-range) and wait for 3+ reconnect backoff cycles — verify the status says "Ring unreachable — reconnecting automatically" (no "charging" assertion).
- **Inferred charging hint**: with the ring on the charger (battery rising), disconnect/reconnect after ≥ 2 descriptor reads — verify "· inferred charging" appears in the reconnect status. With a discharging ring, verify the hint does NOT appear.
- **`liveReadingsStale` unaffected**: HR/SpO₂/steps/temp staleness behavior should be unchanged (30 s monitoring / 360 s idle thresholds still apply).

### DIFF
--- a/ios/OpenRingConn/BLE/RingSession.swift
+++ b/ios/OpenRingConn/BLE/RingSession.swift
@@ -73,6 +73,56 @@ final class RingSession: NSObject {
     /// UI can show a subtle "auto-updating" cue instead of reading as a user measurement.
     private(set) var autoMeasuring = false
 
+    // MARK: Battery freshness (#57)
+    //
+    // Battery % is updated ONLY on 0x10/0x87 descriptor frames (DeviceStatus.battery), which
+    // are solicited by the keepalive every 60–300 s. Using the global `lastFrameAt` (refreshed
+    // on every frame, including 2-s live-HR polls) would let the battery show as "live" for up
+    // to 6 minutes (idleStaleAfter) even though the reading is tens of minutes old during a long
+    // monitoring session. A dedicated per-read timestamp + a tighter 120 s window catches a
+    // silently stale reading after roughly 2 night-keepalive intervals.
+
+    /// Wall-clock of the most recent 0x10/0x87 frame that carried a valid battery % (#57).
+    /// Separate from `lastFrameAt` (updated on every frame) — battery freshness is independent
+    /// of live-HR polling.
+    private(set) var batteryFetchedAt: Date?
+
+    /// True when the last battery reading is old enough to display as stale — i.e. no
+    /// 0x10/0x87 descriptor arrived recently (#57). Tighter than `liveReadingsStale`
+    /// (idleStaleAfter 360 s), which covers all readings. Shows "as of Xm ago" in the
+    /// connection-header battery after `batteryStaleAfter` seconds of silence.
+    var batteryStale: Bool {
+        guard batteryPercent != nil else { return false }   // nothing to call stale yet
+        guard let at = batteryFetchedAt else { return false }
+        return Date().timeIntervalSince(at) > Self.batteryStaleAfter
+    }
+    /// ~2× the tightest keepalive interval (night: 60 s). Battery shows "as of Nm ago" when
+    /// no descriptor has arrived in this window. Daytime keepalive (180–300 s) means the battery
+    /// will accurately report as stale between keepalive firings — that IS correct, the reading
+    /// IS a few minutes old.
+    private static let batteryStaleAfter: TimeInterval = 120
+
+    // MARK: Charging inference (#60)
+    //
+    // The ring's charging byte is NOT on the wire in any confirmed field (PROTOCOL.md §5).
+    // A strictly rising battery % across consecutive 0x10/0x87 frames is the best 🟢 proxy.
+    // This window is private; `inferredCharging` is computed from it via the kit's pure function.
+    // The inference is persisted before session teardown (see `invalidate`) so the connection
+    // card can surface a hint during the reconnect backoff when session == nil.
+
+    /// Rolling battery % readings (oldest→newest, capped) for charging inference (#60).
+    private var batteryTrend: [Int] = []
+    private static let batteryTrendCapacity = 4
+
+    /// True when the last few battery readings are strictly rising — suggesting the ring may
+    /// be charging (🟢 proxy). The UI labels this "inferred" to avoid asserting certainty.
+    /// Never derived from the reconnect count (#60).
+    var inferredCharging: Bool { ChargingInference.inferred(from: batteryTrend) }
+
+    /// UserDefaults key for the last-persisted charging inference (#60). Written before session
+    /// teardown so ContentView can read it during the reconnect-backoff window (session == nil).
+    private static let inferredChargingKey = "battery.inferredCharging"
+
     /// True when frames have stopped for long enough that the live readings (HR/SpO₂/battery/
     /// steps/temp) should read as STALE rather than current (#36). A silently-dropped link keeps
     /// its last values until CoreBluetooth eventually fires `didDisconnect`; this lets the UI
@@ -171,6 +221,10 @@ final class RingSession: NSObject {
     /// reading via `stopLiveMonitoring()` BEFORE this where that matters; `invalidate()` itself is
     /// a pure cancel + detach. Idempotent.
     func invalidate() {
+        // Persist the charging inference BEFORE cancelling tasks (#60): the session is about
+        // to be nil-ed by the scanner; ContentView reads from UserDefaults during the
+        // reconnect-backoff window so the hint stays live while reconnecting.
+        UserDefaults.standard.set(inferredCharging, forKey: Self.inferredChargingKey)
         monitorTask?.cancel(); monitorTask = nil
         keepaliveTask?.cancel(); keepaliveTask = nil
         autoMeasureTask?.cancel(); autoMeasureTask = nil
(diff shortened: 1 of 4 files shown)


perezjuanj:
## 🔍 Adversarial peer review — PR #64 (battery freshness #57 + honest charging copy #60)

Reviewed the full diff plus `RingSession.swift`, `ContentView.swift`, `RingScanner.swift` (read-only), `ReconnectBackoff.swift`, and the new `ChargingInference` kit file/tests. I also built the kit and ran the suite locally.

### Verification I ran
- `swift build` (OpenRingKit): **Build complete** ✅
- `swift test`: **134 tests, 0 failures**, incl. 13 `ChargingInferenceTests` — matches the PR body's claim ✅
- Scope: `git diff --name-only` vs base = exactly the 4 declared files. `RingScanner.swift` is **untouched** (read-only, as required). No edits to measure/auto-measure/not-worn/sleep/empty-state/SpO2 paths ✅

### Fabrication check (repo rule #1) — PASS
No invented protocol value or fake device-state byte. The charging hint is sourced **only** from the labeled battery-rising proxy (`ChargingInference.inferred(from: batteryTrend)`), never from the reconnect count. `statusText` drops the old "or charging" assertion; `ChargingInference.swift` and the RingSession comments explicitly state the charging byte is not on the wire (🔴) and label the result "inferred". This is exactly what #60 asked for.

### Must-fix
None.

### Should-fix
1. **Persisted inference has no recency bound** — `RingSession.swift:226` writes `inferredCharging` to UserDefaults on every `invalidate()`, and `ContentView.swift:536-538` reads it during the stalled window. In the normal disconnect flow this is fine (teardown overwrites with the latest session's value *before* `reconnectStalled` can flip true after 3 attempts). But there's a narrow stale-true path: a session ends with `inferredCharging == true` (ring was charging), the app is killed with **no** subsequent session, then on relaunch the ring is discharging/out-of-range and the reconnect stalls → the UI shows "· inferred charging" from a value that may be hours old. It's softened by the "inferred" label, but consider persisting a timestamp (or the trend itself) and ignoring the hint past, say, one keepalive window.
2. **Magic-string key duplicated** — `RingSession.swift:124` defines `inferredChargingKey = "battery.inferredCharging"`, but `ContentView.swift:537` hard-codes the literal again instead of referencing a shared constant. They match today; a future rename of one silently desyncs them. Promote the key to a shared (e.g. kit) constant or expose it.
3. **2-reading "strictly rising" is a weak proxy** — `ChargingInference.swift` calls `≥ 2` "deliberately conservative", but two rising distinct readings is a low bar that battery-% noise / voltage-sag recovery on a worn ring can satisfy, yielding a false "inferred charging". The `batteryTrend` consecutive-dedup helps with flat noise but not a genuine 2-step rise. Requiring 3 rising readings would meaningfully cut false positives; optional given the hint is labeled inferred. (Note: the falling-then-rising window correctly errs toward *not* asserting charging — good direction.)

### Minor / pre-existing
- `batteryStale` (RingSession.swift:94-98) is a `Date()`-derived computed property, so SwiftUI won't re-render purely from time elapsing — the grey-out / "as of Xm ago" can lag until the next observable change. This is the **same pattern as the existing `liveReadingsStale`**, not a regression introduced here; in practice the keepalive/HR polls drive frequent re-renders.

### What's good
- #57 is correct and on-spec: dedicated `batteryFetchedAt` + a tighter `batteryStaleAfter = 120 s` (within the requested 60–120 s), cleanly separated from `idleStaleAfter` (360 s). `batteryStale` correctly guards both `batteryPercent` and `batteryFetchedAt` nils; `batteryAsOf` re-anchors to `batteryFetchedAt` so a 2-s HR poll no longer keeps the timestamp artificially fresh. Fresh `0x10/0x87` clears staleness. Battery decode (DeviceStatus byte[1], 🟢) is untouched.
- #60 copy is honest: "Ring unreachable — reconnecting automatically", no charging claim from the backoff count; the calm #35 `reconnectStalled` gating is preserved verbatim.
- `ChargingInference` is a clean pure function (mirrors the `ReconnectBackoff` kit pattern), well placed for unit testing, with thorough table-driven tests (empty/single/rising/flat/falling/mixed all covered and verified passing).
- Concurrency is sound: all new state is on the `@MainActor`-isolated `RingSession`; `invalidate()` persists synchronously before cancelling tasks; no new escaping closures, so no retain-cycle risk.

### Verdict
**Approve.** Acceptance for #57 and #60 is met, no fabrication, scope is clean, build/tests verified locally. The three should-fix items (stale persisted bool, duplicated key, weak 2-reading proxy) are quality/honesty refinements, not blockers.

PR #65, opened 2026-06-17
Manual HR/SpO2 measurement: timeout + clear progress/guidance (#55)

### LINKED ISSUE #55: Manual HR/SpO2 measurement: self-terminating timeout + clear progress and guidance
## Problem
A user-initiated "Measure" has no deadline. The poll loop is `while !Task.isCancelled { write(poll); sleep 2s }` with no timeout — only the *auto*-measure path has one (90s). On poor contact / off-wrist / on-charger, a tap spins "Measuring HR…" or "Warming up HR… (8)" forever, with no failure message and no instruction to the user. Two related clarity gaps compound it: (a) the warm-up state surfaces the **raw sentinel byte** (e.g. `(8)`) rather than guidance, so users can't tell a converging read from a stuck one; (b) `livePreparing` (the history-drain that precedes a live read) is set but never shown, so a long backlog drain just reads as a slow "Measuring…".

This is the UX layer on top of #45 (which fixed the on-demand no-op / re-arm and background plumbing). Reference #45; do not duplicate it.

## Proposed fix
- Give the user poll loop a deadline (mirror the auto-measure 90s HR / 45s SpO2 budgets) and, on expiry without a lock, transition to an explicit **failure state** with actionable copy: "Couldn't get a reading — make sure the ring is worn snugly and not on the charger, then hold still."
- Distinguish **"Preparing…"** (drain / `livePreparing`) from **"Measuring…"** in the UI so a backlog sync doesn't masquerade as a stuck measurement.
- Replace the raw "Warming up HR… (8)" with plain copy + a hold-still hint; optionally render the already-collected `liveHRTrend` (capped at 12 samples) so the user sees convergence vs. stall.

## Acceptance criteria
- Tapping Measure on a doffed/charging ring ends within the timeout with an actionable message, never an infinite spinner.
- During a history drain the UI shows "Preparing…"/"Syncing…", not "Measuring…".
- Warm-up copy contains no raw byte and includes a hold-still hint.
- A successful read still shows the locked value as today.

## File pointers
- ios/OpenRingConn/BLE/RingSession.swift:288-292 (user poll loop with no deadline 🟢 confirmed root cause)
- ios/OpenRingConn/BLE/RingSession.swift:424-439 (auto-measure timeout pattern to mirror), :50-52,211 (`livePreparing` set but unread), :33,898-899 (`liveHRTrend` collected, never rendered)
- ios/OpenRingConn/ContentView.swift:180-188 (status surface), :240-248 (`hrWarmupText` shows raw byte2)
- ios/OpenRingConn/VitalsTableView.swift:132-134 (Measure entry)

## Confidence
Live HR decode 🟢; warm-up sentinel / not-locked threshold 🟢 (PROTOCOL.md §5.1). No new protocol facts required — pure client UX + timeout.

---
**Priority:** P1 · **Complexity:** medium · **Recommended impl model:** sonnet · **Actionable now:** yes · **Related:** #45
<sub>Filed from the APK-parity + UX/reliability triage round.</sub>

perezjuanj:
## Summary

- **(a) Self-terminating poll loop (#55a):** The user-poll loop previously had no deadline and could spin forever. It now enforces a per-mode budget (HR 90 s, SpO₂ 45 s — matching the auto-measure budgets). On expiry without a lock, `RingSession.userMeasureFailed` is set with actionable copy. The auto-measure path (`autoMeasureOnce`) is unaffected — it has its own outer deadline and cancels the task externally.
- **(b) Preparing → Measuring distinction (#55b):** `livePreparing` was set during the history drain but never surfaced. ContentView's connection-card header now shows "Preparing…" while the drain runs (`userMeasuring && livePreparing`), then transitions to "Measuring…" once the poll loop starts. VitalsTableView's measure-row time label mirrors the same split ("preparing…" / "measuring…").
- **(c) No raw byte / actionable failure copy (#55c):** "Warming up HR… (8)" (raw sentinel byte exposed) replaced by "Hold still — getting a reading" when warmup frames arrive, "Measuring heart rate…" otherwise. SpO₂ gets its own copy. Timeout failure shown as an inline orange banner (styled like the existing `activationHint`), dismissed automatically when the user taps Measure again.

## New RingSession state (MARK: User measure UX #55)

- `userMeasuring: Bool` — true while the user has hand-started a measure (not auto); drives the deadline and Preparing/Measuring UI distinction
- `userMeasureFailed: Bool` — set on timeout without lock; cleared on retry/rearm/mode-switch/stop
- `userMeasureFailedMessage: String?` — actionable guidance copy for the failure banner

## On-device verification needed

- Tap Measure (HR). Connection card shows **"Preparing…"** while the drain runs, then **"Measuring heart rate…"** or **"Hold still — getting a reading"** — NO raw byte visible.
- With the ring held still, it should lock within ~60 s. With the ring off or loose, wait 90 s — connection card shows the orange **"Reading timed out"** banner; tapping Measure again dismisses it and starts fresh.
- Tap Measure (SpO₂). Same Preparing → Measuring progression; timeout after 45 s if no lock.
- Auto-measure (triggered every 10 min after connect): shows "Auto-measuring…" as before — no change to that path, no failure banner appears for auto.
- Foreground sync: "Updating…" as before — unaffected.

## Kit build / tests

All 121 existing OpenRingKit tests pass — no new tests needed (timeout logic lives in the app-target `RingSession.swift` poll loop which cannot be unit-tested without CoreBluetooth).

Closes #55


perezjuanj:
## 🔍 Adversarial peer review — PR #65 (#55, manual-measure timeout UX)

Reviewed `gh pr diff 65` against the three OWN files. Scope is clean (only RingSession/ContentView/VitalsTableView, within the named regions; auto-measure internals, `statusText`, battery card, sleep, SpO₂ caveat all untouched). No fabrication. There is one real correctness regression on the **success** path.

### 🟥 Must-fix

**1. Zombie session after a *successful* measure outlives its budget — `RingSession.swift`, poll loop (new `if let deadline = pollDeadline …` block + the `if let self, self.userMeasureFailed { stopLiveMonitoring() }` teardown line).**

The deadline `break` is **unconditional** — it fires even when a reading *is* locked:

```swift
if let deadline = pollDeadline, Date() >= deadline {
    let locked = self.liveMode == .hr ? self.liveHR != nil : self.liveSpO2 != nil
    if !locked { self.userMeasureFailed = true; … }
    break                     // <- breaks on success too
}
…
if let self, self.userMeasureFailed { self.stopLiveMonitoring() }   // <- cleanup only on FAILURE
```

So the common path — user taps Measure (HR), it locks in ~30 s, and the screen stays open — at t=90 s breaks the loop with `locked == true`, which means **no cleanup runs**. The poll `Task` returns while leaving `monitoring = true`, `userMeasuring = true`, and `monitorTask` non-nil (a completed Task). Consequences, all reachable without reconnect:

- **Live HR freezes** — polling stopped, but the row/button still read "active"; value goes stale silently.
- **Auto-measure is permanently blocked** — `idleForAutoMeasure` requires `!monitoring` (RingSession.swift:417), so the periodic refresh (the app's "primary tracker" premise) never runs again.
- **`.onChange(monitoring)` → `flushHealth()` never fires** (ContentView.swift:99-100) because `monitoring` never flips to `false`; the locked reading isn't flushed on session end the way the PR's own comment claims.
- **Re-arm / mode-switch become no-ops** — `rearmUserMeasure` and `setLiveMode` both document that they rely on the *existing live poll loop* to keep sending `95 00 00`; that loop is now dead. A re-tap clears `liveHR` and shows "Measuring heart rate…" **forever with no timeout** (the timeout-enforcing loop is gone), and a fresh `startLiveMonitoring` is blocked by `guard monitorTask == nil` (RingSession.swift:206).

Recovery requires a manual Stop tap or a reconnect. Fix: run the teardown on **any** deadline exit (move `stopLiveMonitoring()` out of the `userMeasureFailed` guard), or only `break` when `!locked` (preserves the pre-PR forever-poll on success). The first is cleaner — it also auto-stops a successful measure at its budget and triggers the Health flush.

### 🟧 Should-fix / note

- **Stale deadline on re-arm (minor).** `pollDeadline` is a local captured once after the drain. A user re-tap while a measure is still live (`monitoring == true`, pre-deadline) routes to `rearmUserMeasure`, which does **not** reset that deadline — a late re-arm inherits the original budget and can fail almost immediately. (With the must-fix applied this window is bounded, so low priority.)
- **VitalsTableView scope (low).** The ticket scoped `VitalsTableView.swift` to "only if needed to surface failure"; the change instead uses it for the preparing/measuring split (part b). In-region and harmless, just slightly beyond the stated remit for that file.
- **Test claim is true but irrelevant.** "121 OpenRingKit tests pass" — correct, but every line this PR changes is in the **app target** (`RingSession`/`ContentView`/`VitalsTableView`), which those Kit tests don't touch. The new timeout/teardown logic has zero automated coverage; the success-stays-open-90 s case above specifically needs a manual on-device pass.

### 🟩 What's good

- **No fabrication.** The "…not on the charger…" copy is *user guidance* in a failure message, not a device-state assertion — no invented charging byte, consistent with the repo's #1 rule. Deadlines (HR 90 s / SpO₂ 45 s) mirror the documented auto-measure budgets; no new protocol values.
- **Failure path is correct & self-consistent.** On no-lock timeout it sets `userMeasureFailed` + message, stops monitoring, keeps the banner, and retry/re-arm/mode-switch each clear it. `userMeasureFailed` is deliberately preserved across stop. Retry works because the failure teardown nils `monitorTask`.
- **Preparing → Measuring is well-gated** — correct precedence vs. auto/sync, and `userMeasureInProgress` requires `livePreparing == false` so the two states are distinct.
- **Raw warm-up byte gone**, replaced with honest "Hold still"/"Measuring…" copy. `[weak self]` throughout, no retain cycles; concurrency is fine (`@MainActor` class, Task inherits the actor).

### Verdict

**Request changes** — one must-fix: the unconditional deadline `break` with failure-only cleanup leaves a zombie `monitoring`/`monitorTask` state after a successful measure's budget elapses, freezing live HR and silently disabling auto-measure until reconnect.

perezjuanj:
## ✅ Addressed review

Fixed the blocking teardown bug and the related re-arm footgun. Touched only `ios/OpenRingConn/BLE/RingSession.swift`.

### Must-fix: teardown on ANY deadline exit
The poll loop's deadline `break` was unconditional (fires whether or not a reading locked), but the post-loop teardown only called `stopLiveMonitoring()` when `userMeasureFailed` was true. So in the **common** path — tap Measure, it locks, screen stays open past the 90 s budget — the poll Task ended while leaving `monitoring=true`, `userMeasuring=true`, and a **completed** `monitorTask` non-nil.

Consequences this caused (now fixed): live HR froze, auto-measure was permanently blocked (`idleForAutoMeasure` requires `!monitoring`), `.onChange(monitoring)`→`flushHealth` never fired, re-arm/`setLiveMode` no-op'd, and a fresh `startLiveMonitoring` was blocked by `guard monitorTask == nil`. Recovery needed a manual Stop tap or reconnect.

**Fix:** `stopLiveMonitoring()` (the existing full teardown — clears `monitoring`/`userMeasuring`, nils `monitorTask`, persists the last reading) now runs on **any** deadline exit, inside the deadline branch before `break`. `userMeasureFailed` is set **only** when no reading locked. The dead post-loop conditional teardown is removed. No new teardown path invented; MainActor/concurrency semantics unchanged (the call already ran on the MainActor-inherited poll Task, exactly as the old post-loop call did).

### Should-fix: reset the deadline on re-arm
The deadline was a Task-local captured **once** after the drain, so a re-tap (`rearmUserMeasure`) inherited the original budget and could time out almost immediately. Moved it to a session-held `userMeasureDeadline` (re-read each poll iteration) and reset it on re-arm via a shared `armUserMeasureDeadline()` helper (also used at poll start). Auto path leaves it `nil` (still bounded by `autoMeasureOnce`). Cleared in `stopLiveMonitoring()`/`invalidate()` alongside the other teardown state.

### Build verification
App-target-only change, so `swift build` doesn't exercise it. Verified the app target compiles with the real toolchain (Xcode 26.5):

```
cd ios
xcodegen generate
xcodebuild -scheme OpenRingConn -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build
→ ** BUILD SUCCEEDED **
```

Issue #38, opened 2026-06-15
0x47 PPG pages acked but never decoded — no daytime continuous HR / true HRV / real sleep staging


bluna-ai:
## Problem
To fully replace the official app, the ring's continuous daytime data must be captured. Currently the `0x47` PPG history pages are **acknowledged but discarded**:

- `RingSession.didUpdateValueFor` case `0x47` (`BLE/RingSession.swift:446-451`): acks (`pageAck47`) and feeds the gated epoch session, then `return` — "PPG page — BulkSleep decode TODO (issue #8)."
- `epochDecodingEnabled = false` (`RingSession.swift:83`), so `handlePPGPage`/`handleEndOfHistory` persistence is dead code (`:489-507`).

Consequences for a daily driver:
- **No daytime HR history.** HR only exists from the `0x4c` sleep-vitals epochs (overnight) + manual live measurements. The whole day is blank in Apple Health unless the user manually measures.
- **No true HRV time series**, no daytime SpO2, no real per-epoch data → sleep staging stays a motion+overnight-HR approximation (`SleepStaging.swift` header: "APPROXIMATION, NOT GROUND TRUTH").
- Calories/strain can't be computed accurately without daytime HR.

## Proposed fix
- Prioritize decoding the `0x47` payload (tracked by #8/#9) — this is the single biggest data-completeness unlock.
- Once decoded, flip `epochDecodingEnabled`, route to `Metrics`, persist + HealthKit-write daytime HR/HRV/SpO2.
- This makes "always-on tracker" real instead of "overnight + manual-tap tracker."

Priority: **P1** (data completeness is core to the product goal; blocked on protocol RE)

perezjuanj:
> **Automated triage note (Claude Code).** This ticket is gated on a **physical-ring BLE capture** that an automated coding agent cannot perform, so it is being **deferred** from the current implementation pass rather than ignored. Flipping epochDecodingEnabled and routing 0x47 → daytime HR/HRV requires the payload to be proven first (blocked on #8/#9). The plumbing is in place (pages are acked + fed to EpochSyncSession); only the decode is missing.

perezjuanj:
**Partial resolution: the "no daytime continuous HR" consequence is solved — a different way than this ticket assumed.**

[...]

Issue #119, opened 2026-06-28
Overnight sleep truncated to first fragment: foreground-only drain stalls while app is suspended → early-morning history never drained


perezjuanj:
## Summary
A night gets **truncated at the last successful overnight drain**, losing the back half. Reported by @padawer2 on build 13 (Gen 3, FR05.008): OpenCircuit showed **4h15m** vs Fitbit ground truth **~7h**. This is a **drain-reliability hole**, not a staging error — the back-half epochs were never pulled off the ring.

Source: Discussion #111, diagnostics `opencircuit-diagnostics-2026-06-27T14-35-38Z.txt`.

## Evidence (Randy's 6/26→6/27 night)
Stored summary stops dead at 02:43:
```
2026-06-26 22:27 → 2026-06-27 02:43   asleep 255m (D23/R55/L177/A1)   score 68
```
255m = 4h15m = the "~4 hours" he saw. Real night ran to ~05:30.

Sync activity log — 30-min night cadence works until 02:28, then collapses:
```
22:12 → 02:28   drained every ~30 min, all "ok"     ← night captured fine to 02:43
02:56           FAIL  no frames received
   …no attempts at all for ~3h…                      ← Last background run: —
05:52           FAIL  no frames received
05:56           ok    (one slice)
06:00 / 06:43 / 07:50  FAIL
```
Gap report confirms permanent holes across `02:41 → 05:49` (history NEVER drained). Per-metric cursors stall in the same window (respiratoryRate 02:41, temperature 02:51, hrvSDNN 04:25) while heartRate/spo2 advanced past 07:00.

## Root cause
1. **Foreground-only drain cadence.** `HistoryDrainCadence` fires every ~30 min *while the app is foreground+connected*. `Last background run: —` shows no background drain ran overnight, so when iOS suspended the app after ~02:28 the cadence stopped, and the 02:43→05:30 stretch was never pulled.
2. **Morning re-drain didn't recover the buffer.** The ring buffers for days (per `HistoryDrainCadence` header), so the back half *should* be recoverable at the 05:56 sync — but the gaps persisted. Needs investigation: cursor state after failed fetches, or contention with another app/phone holding the ring.

The night summary is then just "everything up to the last good drain" → truncated at 02:43.

## Not addressed by builds 14/15
- Build 14's only sleep change (`markLeadInWakeOnset`) pushes onset *later* — an over-count fix, opposite direction, and end-of-night only.
[...]

PR #121, opened 2026-07-01
feat(ppg): decode raw PPG (0x47) + BP/SpO2/ECG estimation (#120)

### LINKED ISSUE #120: protocol-re: decode raw PPG (0x47) + downstream BP/SpO2/ECG estimation
Following up on #38 and #8 — I decoded the raw PPG (`0x47`) payload well enough to get continuous per-sample channel data off the ring, and built a few things on top of it:

- **Raw PPG protocol decode**: identified channel layout and framing for `0x47` (`desktop/analyze_ppg_13.py`, `desktop/decode_4d_4e.py`, `desktop/identify_channels.py`, `desktop/probe_ppg_raw.py`, `desktop/probe_ppg_params.py`), plus live streaming/capture tooling (`desktop/stream_ppg_13.py`, `desktop/capture_ppg.py`, `desktop/ppg_pipeline.py`, `desktop/ppg_feature_extractor.py`).
- **Blood pressure estimation** via pulse-transit-time (`desktop/ptt_calculator.py`, `desktop/bp_estimator.py`, `desktop/bp_calibration.py`) — calibrated against a cuff reference.
- **SpO2 calibration** (`desktop/spo2_calibrator.py`) and **ECG R-peak detection** (`desktop/ecg_rpeaks.py`) from the same raw PPG stream.
- A matching iOS layer: calibration session UI, history-sync diagnostics views, and an auth module, with tests.

BP and ECG are outside what's currently on the roadmap (Phase 5 lists HRV/stress/strain/sleep score), so I split them out as optional/bonus rather than assuming they belong in the core interoperability scope — happy to gate them behind a flag or keep them as a separate module if that's a better fit.

I rebased this on top of build 17 (merged in the onboarding/diagnostics/sleep-confidence work) — Swift package builds clean, 582/582 tests pass. Opening a PR shortly that references this issue.

pravinsail:
Closes #120, which also addresses #38 and #8.

## What this adds

- **Raw PPG protocol decode** (`0x47`): channel layout identification, live streaming/capture, and feature extraction — the piece #38 and #8 were blocked on.
- **Blood pressure estimation** from pulse-transit-time, with a calibration workflow against a cuff reference.
- **SpO2 calibration** and **ECG R-peak detection** from the same raw PPG stream.
- Matching iOS UI (calibration session flow, history-sync diagnostics) and tests.

BP/ECG are beyond the current roadmap's stated scope (interoperability bridge to Apple Health, not a full vitals-estimation app) — flagging that explicitly in case you'd rather these land as an optional module, or not at all. The PPG decode itself (which unblocks #38/#8) is the part I'd consider core.

## Verification

- Rebased onto current `master` (build 17) — merged cleanly except 2 conflicts in `RingSession.swift` and `ContentView.swift`, both resolved by keeping both sides' independent fixes (not overwriting either).
- `swift build` — clean.
- `swift test` — 582/582 passing.
- No real capture data, device identifiers, or local paths included — this PR only carries decoded findings/tooling, consistent with `desktop/captures/` staying gitignored.

Happy to split this into smaller PRs (PPG decode alone vs. BP/ECG on top) if that's easier to review.


ios/OpenCircuit/Observability/HistorySyncEvidenceStore.swift
```diff
@@ -0,0 +1,49 @@
+import Foundation
+import OpenCircuitKit
+
+/// One persisted history-sync evidence bundle. Kept separate from the human-readable activity log:
+/// this is a machine-oriented breadcrumb for "why did sleep not land?" incidents.
+struct HistorySyncEvidence: Codable, Identifiable, Equatable {
```
perezjuanj: 🔴 **Blocking — build-breaker.** This declares `struct HistorySyncEvidence` a second time in the same app-target module (canonical one is `ObservabilityStore.swift:37`, used by `observability.recordHistorySyncEvidence(...)`). Two top-level types with the same name in one module → the app target does not compile. The `HistorySyncEvidenceStore` type here is referenced nowhere and the struct body is field-identical, so this whole file is dead

*Fix: delete this file.** Done on branch `fix/pr121-blocking-fixes`; `xcodebuild` then reports BUILD SUCCEEDED.

The Kit's 582 passing tests miss this — the duplicate is in the app target, not the `OpenCircuitKit` package.

ios/OpenCircuit/BLE/RingSession.swift
```diff
@@ -1348,16 +1753,137 @@ final class RingSession: NSObject {
...
+    /// Ask the ring for a fresh device-status descriptor (`0x10`/`0x87`) once the link is idle.
+    /// The descriptor is the source for the "doesn't need overnight staging" metrics:
+    /// today's step counter, battery %, charging state, case battery, and any live-only skin-temp
+    /// snapshot. We try the proven command families in increasing strength: `fetch`, then
+    /// `status0`→`fetch`, then `d0`→`fetch`. If the ring still stays quiet we leave the request
+    /// queued so the next idle transition can retry.
+    private func maybeRequestDeviceStatusRefresh(reason: String) {
+        guard pendingDeviceStatusRefresh, postSyncStatusTask == nil else { return }
+        guard ready else { return }
+        guard !syncing, !monitoring, !livePreparing, !workoutHolding, !calibrationCapturing else {
```
perezjuanj, line 1852: 🔴 **Blocking — overnight sleep-drain regression.** `maybeRequestDeviceStatusRefresh` fires a bare `0x07` `fetch` (plus an escalation ladder of up to 3 more just below) with no `!isInSleepWindow` guard. It is armed by `scheduleDeviceStatusRefresh("post-sync")` at the end of **every** `finalizeSync` — including the cadenced *overnight* drain — so on the next idle keepalive tick it fires `0x07` **inside the sleep window** and walks the ring's single history pointer, skimming the `0x87` header while discarding the `0x4c` sleep-vitals page.

That is the exact pointer-walk the build-17 quiet-drain fix removed (the night keepalive deliberately uses `0xD0` instead of `fetch`). All three of `maybeRequestDeviceStatusRefresh` / `scheduleDeviceStatusRefresh` / `pendingDeviceStatusRefresh` are **new in this PR**.

**Fix: add a `!isInSleepWindow` guard here** so the refresh stays queued (`pendingDeviceStatusRefresh`) and fires in the morning. Done on branch `fix/pr121-blocking-fixes`.

perezjuanj (changes requested):
## Review — PR #121: PPG (0x47) decode + BP/SpO2/ECG estimation

**Verdict: 🔴 Request changes.** Two blocking issues (one build-breaker, one fragile-area regression), and the PR should be split before any of it lands.

Reviewed with a multi-agent fan-out (8 file-group finders + adversarial verification); the two blocking findings were then verified by hand against the code. **I've prepared both blocking fixes on branch `fix/pr121-blocking-fixes`** (see inline comments).

### The scope problem (read first)

The title says "PPG decode + BP/ECG estimation," but a `Merge branch 'local-snapshot-base'` commit swept **~2,600 lines of unrelated shipping iOS work** into it.

| Area | Lines | Matches title? |
|---|---|---|
| `desktop/` RE tooling (25 files) | +6,671 | ✅ throwaway, honestly labeled |
| `docs/` PPG/vitals handoffs | +1,479 | ✅ |
| **`ios/` shipping app** (56 files) | **+6,199 / −617** | ❌ **majority is NOT PPG** |

Only ~4 shipping files (+1,468) are actually PPG. The rest — `LocalStore.swift` (+265, a new sync-cursor repair/plausibility engine, **zero PPG**), ~618 non-PPG lines in `RingSession.swift`, new trends/export/analytics/HealthKit-history code — lands squarely in the **fragile overnight-drain / sync-cursor area** with a documented history of user-facing regressions (self-contention, shrinking sleep, the 0x07 pointer-walk). A reviewer scanning "for PPG" won't scrutinize a sync-cursor rewrite — which is how the past regressions shipped.

**Recommendation — split into 3 PRs:** (1) `desktop/` + `docs/` RE tooling (low-risk, unblocks #38/#8); (2) the BP/ECG/SpO2 calibration bridge (the titled feature — keep it flagged research-only, as you already do); (3) the swept-in shipping iOS work, which **must get its own focused sync-cursor/sleep-drain review**. Whichever way it lands, **squash-merge** — the pre-redaction commit `97a7803` still carries the personal paths/identifiers that `310a383` cleaned up.

### 🔴 Blocking

**B1 — The app does not compile: duplicate top-level `struct HistorySyncEvidence`.**
Declared in two files in the same app-target module: `Observability/ObservabilityStore.swift:37` (canonical — used by `observability.recordHistorySyncEvidence(...)`) and the brand-new `Observability/HistorySyncEvidenceStore.swift:6`. Invalid redeclaration → build error. The new `HistorySyncEvidenceStore` type is referenced nowhere and the struct bodies are field-identical.
→ **Fix: delete `HistorySyncEvidenceStore.swift`** (done on the fix branch).
*Note: the Kit's 582/582 passing tests don't catch this — the duplicate is in the app target, not the `OpenCircuitKit` package. This PR has almost certainly never been built on-device.*

**B2 — Re-introduces the overnight `0x07` history-pointer walk that the build-17 quiet-drain fix removed** (`RingSession.swift`). Verified end-to-end:
- The night keepalive correctly uses `0xD0 statusQuery` instead of `fetch` (build-17 fix survives). **But** the PR adds `maybeRequestDeviceStatusRefresh("idle keepalive")` at the top of every tick with **no `!isInSleepWindow` guard**, and it fires bare `Command.fetch` (`0x07`) plus an escalation ladder of up to 3 more.
- `scheduleDeviceStatusRefresh("post-sync")` fires at the end of **every** `finalizeSync`, including the cadenced *overnight* drain → arms the refresh → next idle tick fires `0x07` **inside the sleep window**.
- All three symbols are **new in this PR** (absent on master); master's keepalive had no top-of-tick refresh. By the code's own comment, `0x07` is "fetch NEXT history record … walks the pointer," skimming the `0x87` header while discarding the `0x4c` sleep-vitals page.
→ **Fix: add `!isInSleepWindow` to `maybeRequestDeviceStatusRefresh` so overnight refresh waits for morning** (done on the fix branch; keeps the request queued via `pendingDeviceStatusRefresh`).

### 🟡 Should-fix (verified real)

- **Calibration reports "uploaded" even when the ground-truth labels silently fail** (`CalibrationSupport.swift:415-418,502`) — `try?` / 404-as-success means an **unlabeled session can feed the BP model** while the UI shows success.
- **StepAccumulator double-counts steps into Apple Health on a same-day reinstall/fresh-pair** (`StepAccumulator.swift:53-55` → `HealthKitWriter.swift:171-181`) — full-counter credit + HealthKit retaining prior-install samples.
- **Workout HR written to Health twice** (`WorkoutSessionManager.swift`) — `finalize` mirrors workout HR into `LocalStore` (→ standalone `.heartRate`) on top of the `HKWorkoutBuilder` HR write.
- **False-confidence regression test** — `SyncCursorPlausibilityTests.testOutOfBandHeartRateDoesNotPoisonCursor`: the garbage sample is timestamped *behind* the legit one, so it passes whether or not the bug exists.
- **Coverage gap** — no test exercises the `isInSleepWindow` self-contention gate despite the +763-line drain/cursor rewrite; the BP→HealthKit write and the ~575-line calibration flow ship untested. Extracting the sleep-window gate into an injectable pure helper would have caught B2.

### 🟡 Data-safety / hygiene (mostly clean — good work)

Redaction is largely done: no `/Users/` paths, no real MAC (placeholder only), serial redacted, no keys. Residual low items: `ios/LICENSE:3` names an individual vs the org brand used everywhere else; a host-local CBPeripheral UUID + a few of the dev's own spot vitals remain in two docs; `ios/scripts/monitor.sh:25` hardcodes the **wrong bundle id** (`dtrh.projectcircuit.com`) so its default mode is broken; `ppg-stage1-handoff` marks PPG channel identity "Confirmed" while the LED test is still 🔲 Pending in the sibling log — downgrade to 🟡.

### ✅ Credit

- **No fabricated health values** anywhere — SpO2/BP/HR all trace to real signal; BP→Health is opt-in (default OFF), cuff-calibrated, metadata-tagged.
- **Merge-conflict resolution is sound** — the build-17 quiet-drain fix, the 14h night-scoping cap, and both sides of `ContentView` survived intact; no dropped/duplicated blocks, no leftover markers.
- New sleep/sync regression tests added; PPG/BP decode kept **diagnostic-only** (never reaches HealthKit); PROTOCOL.md additions properly 🟡-tagged; **582/582 Kit tests pass**.

### Nits (low)

Desktop DSP: ECG R-peak abs-argmax can lock onto the S-wave trough (±20–40 ms PTT jitter); SpO2 constant mismatch (`analyze_ppg_13.py` B=25 vs pipeline B=19.05); BP POSTed with no physiological clamp; FFT sub-harmonic dealiasing can halve HR at a permissive 30% threshold. iOS: dead `flushPassive/ActiveCalories`; stray `print()` in shipping `flushHealth`; `fetchMostRecentECG` runs off-MainActor; `distanceWalkingRunning` silently cut ~3× (0.248 m/step) — worth one ground-truth check.

---

**Suggested path:** carve out `desktop/` + `docs/` as its own clean PR (unblocks #38/#8), land B1+B2 (branch `fix/pr121-blocking-fixes`), then send the shipping-iOS half up for a dedicated sync-cursor review.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

pravinsail:
Pushed a fix for both blocking findings (`2815000`):

- **B1** — deleted `Observability/HistorySyncEvidenceStore.swift`, the unreferenced duplicate of `HistorySyncEvidence` that was breaking the app-target build.
- **B2** — added a `!isInSleepWindow` guard to `maybeRequestDeviceStatusRefresh` in `RingSession.swift`, so the idle-keepalive status refresh can no longer fire a `0x07` fetch inside the overnight window. The request now stays queued (`pendingDeviceStatusRefresh`) and only resolves once the sleep window ends, so it doesn't re-open the history-pointer-walk regression.

Scoped this to just the two blocking items for now rather than folding in the should-fix/hygiene list or the desktop+docs/shipping-iOS split — will follow up separately on the scope split you suggested. Ready for re-review on the blocking pieces.

perezjuanj:
Merged onto current `master` via local integration. Heads-up on the reconciliation: your branch was based on build-17, and #119 (background sync) had since rewritten the same overnight-drain + step-write paths, so this no longer fast-forwarded. I kept both sides — #119's overnight-quiet keepalive alongside your `maybeRequestDeviceStatusRefresh` (behind the `!isInSleepWindow` guard from the B2 fix so it can't walk the 0x07 pointer overnight), and your per-snapshot `StoredStepSample` step→Health path (it subsumes the recent single-window fix). Verified: app target builds + OpenCircuitKit 602/602. Thanks for the PPG decode work — that's the piece #38/#8 were blocked on. 🎉

Issue #85, opened 2026-06-17
feat: skin-temp & fever anomaly notifications (wire the 5 temp flags + fever)


perezjuanj:
## Problem / Opportunity
Fills the seam between the skin-temp and alerts tickets: the temp baseline ticket computes the 5 temperature anomaly flags, the vitals-anomaly ticket computes suspected fever — but the health-alerts ticket scopes ONLY high-HR + low-SpO2. Without this ticket, NO temperature or fever notification ever fires, despite the catalog enums and "exhibiting suspected fever symptoms" copy.

## Evidence from APK
- Catalog enums `skinTempAbnormalRise/Drop` (0x12/0x13), `skinTempFluctuationRise/Drop` (0x10/0x11), `feverAbnormal` (0x14).
- "exhibiting suspected fever symptoms" (pp.txt:48241); `keyWBVitalSignFever`.

## Scope
- Local `UNUserNotification`s fired from the 5 temp anomaly flags (owned by the temp baseline ticket) + the suspected-fever flag (owned by the vitals-anomaly ticket).
- Respect the SAME shared DND/quiet-window + anti-spam engine as the other notification tickets.

## Acceptance criteria
- Each temp/fever flag produces a notification on real threshold crossings; respects DND + backoff; tests for flag→notification routing.

## Blockers / Deps
- Depends on the skin-temp baseline ticket (supplies temp flags) AND the vitals-anomaly ticket (supplies fever flag). Both are Wave 1. Carry medical-disclaimer copy.

---
**Status:** `implementable-now` · **Wave 1** · **Complexity:** M · **Recommended model:** sonnet

**Depends on:** #69, #72

<sub>Auto-filed from RingConn APK gap analysis (decompiled v3.2.1).</sub>

perezjuanj:
Implemented in Wave 1 and merged to master via #98 (merge commit 04df682). Closing as addressed.

PR #122, opened 2026-07-03
fix: dedupe temp health alerts per night

bluna-ai:
## Summary

- Add per-night dedupe for skin-temperature/fever health notifications so the same overnight summary does not re-alert after every sync.
- Reserve health-alert notifications synchronously before notification authorization awaits, closing the foreground/sync-complete double-post race.
- Add focused coverage for same-night suppression, first-time pass-through, next-night re-arm, and stale older-night suppression.

## Verification

- `git pull --ff-only`
- `brew install xcodegen`
- `softwareupdate --install 'Command Line Tools for Xcode 26.6-26.6'`
- `xcodegen generate`
- `swift build` from `ios/OpenCircuitKit`
- focused compiled Swift assertion executable linked against the built `OpenCircuitKit` object files
- `git diff --check`

## Notes

- `swift test` still cannot run on this host because the Apple Command Line Tools install does not provide `XCTest` (`no such module 'XCTest'`).
- `xcodebuild` still cannot run because full Xcode is not installed/selected; this host only has `/Library/Developer/CommandLineTools`.

### DIFF
--- a/ios/OpenCircuit/HealthNotificationCenter.swift
+++ b/ios/OpenCircuit/HealthNotificationCenter.swift
@@ -136,6 +136,28 @@ struct HealthNotificationStore {
         for n in notifs { raw[n.rawValue] = now.timeIntervalSince1970 }
         defaults.set(raw, forKey: Self.key)
     }
+
+    // Per-night ledger for the skin-temp/fever notifications (#85). Separate from `lastFired` (the
+    // rolling anti-spam backoff) because these flags describe ONE overnight summary and must fire at
+    // most once per night regardless of how many syncs land that day — see
+    // `TempFeverNotifications.freshForNight`. Stores each flag's already-notified night start-of-day.
+    private static let nightKey = "alerts.health.lastNight"   // [HealthNotification.rawValue: nightEpoch]
+
+    func lastNotifiedNight() -> [HealthNotification: Date] {
+        let raw = defaults.dictionary(forKey: Self.nightKey) as? [String: Double] ?? [:]
+        var out: [HealthNotification: Date] = [:]
+        for (k, v) in raw where v > 0 {
+            if let n = HealthNotification(rawValue: k) { out[n] = Date(timeIntervalSince1970: v) }
+        }
+        return out
+    }
+
+    func markNight(_ notifs: [HealthNotification], night: Date) {
+        guard !notifs.isEmpty else { return }
+        var raw = defaults.dictionary(forKey: Self.nightKey) as? [String: Double] ?? [:]
+        for n in notifs { raw[n.rawValue] = night.timeIntervalSince1970 }
+        defaults.set(raw, forKey: Self.nightKey)
+    }
 }
 
 // MARK: - The engine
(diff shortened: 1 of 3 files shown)


bluna-ai:
Verification after full Xcode install:\n\n- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test --package-path ios/OpenCircuitKit` passed: 552 tests, 0 failures.\n- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcodebuild -project ios/OpenCircuit.xcodeproj -scheme OpenCircuit -configuration Debug -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build` passed: BUILD SUCCEEDED.\n\nNote: system `xcode-select` still points at Command Line Tools because switching it requires sudo/password, so these checks used `DEVELOPER_DIR` explicitly.

perezjuanj:
## Review

**Verdict: Approve with one recommended change.** The core fix is correct and solves a real user-facing bug. One minor regression is worth addressing before merge (or making an explicit decision on); everything else is nits.

### What this gets right

- **The headline fix is correct and valuable.** The per-night ledger (`lastNotifiedNight`/`markNight`, key `alerts.health.lastNight`) genuinely stops the #85 temp/fever alerts from re-firing every ~2h all day. The 2h backoff is shorter than a day, so the same "skin temperature dropped" alert re-raised on every sync. `freshForNight`'s `night > last` (strictly-newer) predicate correctly handles same-night suppress, next-night re-arm, and stale-older-night suppress.
- **The concurrency reasoning is sound.** There is no `await` anywhere between the `lastFired` read and the two marks, so on the `@MainActor` the synchronous prefix serializes, and moving `markFired` before the `ensureAuthorized()` suspension really does close the double-post window.
- **Good hygiene:** pure logic (`freshForNight`) extracted to `OpenCircuitKit` and unit-tested; the change is contained (`tempFeverCandidates` is private, so the new tuple return breaks no external caller); comments match the file's style.

### Findings worth acting on

**1. `markNight` consumes a night even when the notification was never delivered (minor, recommend fixing)**
`HealthNotificationCenter.swift:243-245`

To win the race, both `markFired` and `markNight` now run before `guard await ensureAuthorized()`. If auth is denied (or the first-run prompt is declined, or the background BGTask path hits `.notDetermined` and can't prompt), the night is recorded as notified but nothing is posted. Unlike the 2h `markFired` backoff, the night ledger has no time-based self-heal: `freshForNight` only re-arms on a strictly newer night, so a real fever/skin-temp flag computed while unauthorized is silently swallowed for that night if the user enables notifications later the same day. (`try? await center.add(...)` swallowing a failure has the same effect.)

Clean fix: keep `markFired` synchronous (it's the load-bearing race guard, since a second `evaluate()` is then gated out by the 2h backoff anyway), but move `markNight` to after `ensureAuthorized()` succeeds. That reopens no meaningful double-post race (the synchronous `markFired` already suppresses the concurrent temp post via the backoff), and the ledger stops claiming nights it never delivered. The added comment currently only explains the race motivation; it should acknowledge this trade-off either way.

Worth noting the race being fixed is itself low-impact: `post` uses a stable identifier `alerts.health.\(n.rawValue)`, so a concurrent double-post largely coalesces into a single refreshed banner. More reason not to trade it for a new regression when splitting the marks costs nothing.

**2. `isTempFever`'s `default: return false` is a latent footgun (minor)**
`HealthNotificationCenter.swift:250-258`

The temp/fever set is now encoded in three places, and this is the only one that closes with `default:` instead of an exhaustive switch. A future skin-temp case added to `TempFeverNotifications.notifications(...)` would silently fall into `default`, get dropped by `fire.filter(Self.isTempFever)`, never be recorded in the ledger, and regress right back to the every-2h re-fire this PR exists to fix, with no compile error. The sibling `appendsDisclaimer(...)` a few lines down already uses the compiler-safe pattern (exhaustive, no `default`).

Fix: drop the `default` and enumerate the seven non-temp cases, or (better) hoist the temp/fever set to one `Set<HealthNotification>` in `OpenCircuitKit` referenced by `notifications`, `isTempFever`, and `appendsDisclaimer`.

### Nits (optional)

- **Timezone-unstable night key** (`HealthAlerts.swift:239`): `markNight` persists `startOfDay(...).timeIntervalSince1970` and `freshForNight` compares a freshly recomputed instant. DST fall-back does not reproduce (Swift's `startOfDay` resolves the offset at the target date), but westward travel (offset decreasing) between two syncs of the same night can shift the instant later and re-fire the duplicate. Rare. To be airtight, compare a timezone-stable `yyyymmdd` integer day key.
- **Race fix applied to `evaluate()` only** (`:352-354`, `:366-368`): `evaluateReminders()` and `postChargingComplete()` still mark after the await. Pre-existing, those methods post disjoint notification sets, and stable identifiers coalesce, so not a defect, just a consistency gap now that the reserve-before-await rationale exists. Fine to leave with a one-line note.
- **Race fix has no test** (`:244-246`): validated by reasoning only. An app-target test bundle does exist (`ios/OpenCircuitTests/`, `@testable import OpenCircuit`), so a call-order spy is feasible if you want a regression guard, but the pure `freshForNight` coverage is the right primary bet.

### Checked and cleared

Migration one-time re-fire (bounded, strictly better than master), `markNight`/`lastNotifiedNight` round-trip "untestable" (false: app-target test bundle exists, and it's symmetric with the pre-existing untested `markFired`), `freshForNight` precondition (caller honors start-of-day on both sides), `markNight`'s `night:` label vs `at:` (intentional and correct), and the `tempNight`/`night` "redundancy" (idiomatic escape-carrier). None are real.

bluna-ai:
Addressed the review.

**1. `markNight` after auth** (`HealthNotificationCenter.swift:245-247`) — `markFired` stays synchronously *before* `ensureAuthorized()` (the double-post race guard, backed by the 2h backoff); `markNight` now runs only *after* auth succeeds, so a denied-auth night is no longer silently claimed and swallowed for the day. The block comment now records the auth trade-off.

**2. `isTempFever` footgun** (`HealthNotificationCenter.swift`) — hoisted the temp/fever set to a single `TempFeverNotifications.notificationSet` in `OpenCircuitKit`. `isTempFever` is now `notificationSet.contains(n)` (no `default:`), and `appendsDisclaimer` references the same set while staying exhaustive (no `default`), so a new skin-temp case can't silently regress to every-2h re-fire.

**Nits**
- Night ledger now keys on a timezone-stable `yyyymmdd` `Int` (`TempFeverNotifications.dayKey`) instead of `startOfDay(...).timeIntervalSince1970`, so westward travel between syncs of the same night can't shift the instant and re-fire. Bounded one-time re-arm on upgrade from the old epoch format. Added a regression test.
- Updated the `markFired`/`markNight` comment to explain the auth trade-off alongside the race motivation.

**Verification** (`DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer`)
- `swift test --package-path ios/OpenCircuitKit` → **553 tests, 0 failures** (was 552; +1 timezone-stability regression test).
- `xcodebuild -project ios/OpenCircuit.xcodeproj -scheme OpenCircuit -configuration Debug -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build` → **BUILD SUCCEEDED**.

PR #123, opened 2026-07-03
fix: narrow HealthKit step sample window

bluna-ai:
## Summary

- Change HealthKit step writes from an all-day `startOfDay -> now` sample to a narrow observation window near the latest ring step update.
- Keep the existing `healthWrittenSteps` delta watermark, so repeated syncs still avoid re-writing the same OpenCircuit steps.
- Apply the same narrowed interval to the derived step-distance sample written in the same HealthKit batch.

## Why

OpenCircuit was writing a pending ring step delta as one additive `.stepCount` sample covering the whole day. Apple Health prioritizes overlapping sources for similar data, so that all-day OpenCircuit sample can shadow Apple Watch/iPhone steps for the entire day and make the Health app show only ring steps after sync.

The ring does not provide per-step timestamps, so this patch writes the delta over `observedAt - 60s ... observedAt` using the existing `StoredDaily.updatedAt` timestamp from the latest ring step observation.

## Verification

- `xcodegen generate`
- `swift build` from `ios/OpenCircuitKit`
- `git diff --check`

## Not Run / Blocked

- `swift test --package-path ios/OpenCircuitKit` still fails on this host because `XCTest` is unavailable in the installed Command Line Tools environment.
- `xcrun --sdk iphoneos --show-sdk-path` fails because the iOS SDK/full Xcode is not installed, so the iOS app target could not be compiled locally.

### DIFF
--- a/ios/OpenCircuit/Health/HealthKitWriter.swift
+++ b/ios/OpenCircuit/Health/HealthKitWriter.swift
@@ -156,22 +156,27 @@ final class HealthKitWriter {
         let profile = Self.storedUserProfile()
 
         // Steps + distance estimate (#81): write both atomically from the same pending delta.
-        // HealthKit SUMS stepCount / distanceWalkingRunning, so writing the delta lands the
-        // day's running total without re-adding on every sync. Distance is an ESTIMATE
-        // (steps × height-based stride — not GPS) and is labeled as such in HealthKit metadata.
+        // HealthKit stepCount is cumulative, but Apple Health resolves overlapping sources by
+        // priority. Writing the ring delta as an all-day sample can shadow Watch/iPhone steps for
+        // that whole day, so the delta is saved over a narrow window near the latest observation.
+        // Distance is an ESTIMATE (steps × height-based stride — not GPS) and is labeled as such.
         // To avoid double counting against a recorded foot-based workout's GPS distance (which
         // also writes .distanceWalkingRunning), the estimate is netted by any uncredited workout
         // GPS distance for today — preferring the accurate GPS measurement (#).
-        if let delta = try? store.pendingStepDelta(), delta > 0 {
+        if let pendingSteps = try? store.pendingStepWrite() {
             let now = Date()
+            let delta = pendingSteps.delta
             let startOfDay = Calendar.current.startOfDay(for: now)
+            let observedAt = Swift.min(Swift.max(pendingSteps.observedAt, startOfDay), now)
+            let writeStart = Swift.max(startOfDay, observedAt.addingTimeInterval(-60))
+            let writeEnd = observedAt
             let rawDistanceM = DistanceEstimate.meters(steps: delta, profile: profile)
             let (netDistanceM, gpsReduction) = Self.netDistanceEstimate(rawDistanceM, day: startOfDay)
             var toWrite: [QuantitySample] = [
-                QuantitySample(kind: .steps, start: startOfDay, end: now, value: Double(delta))
+                QuantitySample(kind: .steps, start: writeStart, end: writeEnd, value: Double(delta))
             ]
             if netDistanceM > 0 {
-                toWrite.append(QuantitySample(kind: .distance, start: startOfDay, end: now, value: netDistanceM))
+                toWrite.append(QuantitySample(kind: .distance, start: writeStart, end: writeEnd, value: netDistanceM))
             }
             do {
                 try await write(toWrite)
--- a/ios/OpenCircuit/Store/LocalStore.swift
+++ b/ios/OpenCircuit/Store/LocalStore.swift
@@ -708,6 +708,18 @@ struct LocalStore {
         return max(row.steps - row.healthWrittenSteps, 0)
     }
 
+    /// Pending step delta plus the most recent ring-step observation timestamp. HealthKit step
+    /// samples should use a narrow observation window, not an all-day range, otherwise Apple
+    /// Health's source-priority handling can let OpenCircuit shadow Watch/iPhone steps all day.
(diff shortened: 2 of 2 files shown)


bluna-ai:
Verification after full Xcode install:\n\n- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test --package-path ios/OpenCircuitKit` passed: 550 tests, 0 failures.\n- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcodebuild -project ios/OpenCircuit.xcodeproj -scheme OpenCircuit -configuration Debug -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build` passed: BUILD SUCCEEDED.\n\nNote: system `xcode-select` still points at Command Line Tools because switching it requires sudo/password, so these checks used `DEVELOPER_DIR` explicitly.

perezjuanj:
## Review

@bluna-ai thanks for this. The diagnosis is right and the code itself is clean, but I think the chosen fix probably does not achieve its goal, and may replace the reported bug with a worse one. I would hold this for on-device verification before merge. Details below.

### The main concern: narrowing an additive sample likely turns shadowing into double-counting (please verify on-device)

`HealthKitWriter.swift:166-180`

The premise is correct: Apple Health resolves overlapping step sources by priority per time interval (highest-priority source wins an interval; non-overlapping intervals are summed across sources). But our `.stepCount` sample is additive and its value is the full `ring_total` (as deltas that sum via the `advanceStepsWritten` watermark). Narrowing the interval does not change the value, only how much of the day it overlaps:

- Before (all-day `[startOfDay, now]`): the ring sample overlaps the whole day. If the ring source ranks higher, Health shows `ring_total` and hides the phone/Watch entirely. That is exactly the reported "Health shows only ring steps." One plausible total, no double count.
- After (~60s windows): the ring only "wins" its scattered `N x 60s` flush windows; the phone/Watch shows through for the other ~24h. The displayed daily total becomes roughly `phone_or_watch_total + ring_total`, i.e. close to double, for any user who also carries an iPhone (the iPhone motion source counts steps for almost everyone, not just Watch owners).

So the window change trades an all-day shadow (a single, plausible number) for a likely daily over-count in the exact scenario the PR targets. A window width cannot make an additive sample stop double-counting while also letting the other source show through: those two goals are in direct tension. The real lever is source priority / whether we should write ring steps at all when an authoritative source is present.

This is medium-confidence: it rests on Apple Health's cross-source dedup model and I could not reproduce it in this environment, so please confirm on-device before merging:
- an iPhone-only account (no Watch): does the Health daily step total after a ring sync read roughly `phone + ring` (double) or the correct value?
- a Watch + ring account: same check.

If it double-counts, the fix likely needs to be one of: suppress ring step writes when an authoritative source (iPhone/Watch) already covers the day, or gate on OpenCircuit being the sole step source, rather than adjusting the window. Worth documenting the priority dependence either way.

### Nits (none blocking on their own)

1. Dead method + duplication + stale doc (`LocalStore.swift:704-725`). `pendingStepDelta(day:)` now has no non-comment callers (its only caller moved to `pendingStepWrite`), and `pendingStepWrite` re-implements its exact `max(row.steps - row.healthWrittenSteps, 0)` formula verbatim. The `advanceStepsWritten` doc still says "so the next `pendingStepDelta` is correct" (line 725), which is now the wrong method name. Either delete `pendingStepDelta` and fix that comment, or have `pendingStepWrite` delegate to it so the delta math lives in one place.

2. Magic constant + degenerate window (`HealthKitWriter.swift:170-172`). The `60` is a bare literal, worth a named `stepSampleWindow: TimeInterval = 60`. Separately, when `observedAt` clamps down to `startOfDay` (reachable only via a `Date.distantPast` `updatedAt` migration remnant for today's row), the window collapses to `[startOfDay, startOfDay]`, a zero-duration sample pinned to midnight. HealthKit accepts start == end so it is harmless, but consider flooring the window to > 0 or noting that an instantaneous sample is acceptable.

3. Big-delta spike on rare flushes (`HealthKitWriter.swift:170-176`). The whole since-last-write delta is written across <= 60s at `observedAt`. When flush runs frequently this scatters into many small windows (fine); on a first-morning or long-gap sync a multi-thousand-step delta lands in <= 60s (a >100 steps/s bar in the hourly chart). The daily total and day attribution stay correct and nothing consumes a step "rate", so this is cosmetic, but worth a one-line comment acknowledging the trade.

4. No test for the clamp (`HealthKitWriter.swift:170-172`). The observedAt/writeStart/writeEnd clamp is the only real behavior change and has zero coverage. It would fit the repo's "pure logic in the kit, tested" convention (like `StepAccumulator` / `DistanceEstimate`) to extract a free function, e.g. `StepWriteWindow.range(observedAt:startOfDay:now:) -> (start: Date, end: Date)` in OpenCircuitKit, and cover the zero-length, distantPast-underflow, future-overflow, and normal cases.

### Checked and cleared

For the record, these were considered and are fine: `writeStart <= writeEnd` holds for all inputs (window length always in `[0s, 60s]`, no invalid interval reaches HealthKit); the zero-length edge is harmless and effectively unreachable on the normal path; the cross-midnight double-`Date()` quirk is pre-existing and not worsened; `row.updatedAt` is a sound anchor (set by `addDailySteps`, not bumped by `advanceStepsWritten`, so the window does not chase the write time); the reworded comment is an accuracy improvement, not a regression; and the after-state compiles (`Swift.min`/`Swift.max` is cosmetic, not a name collision).

### Verdict

Diagnosis correct, implementation clean, but hold for the on-device double-count check above. If Health does double-count with a concurrent source, this needs the priority/gating approach rather than a narrower window. If it does not (priority happens to favor the other source in practice), then the nits are all that remain.

PR #125, opened 2026-07-04
fix: restore on-demand vitals controls

bluna-ai:
## What Problem This Solves

PR #121 removed the home-screen on-demand Heart Rate and SpO2 controls while keeping the underlying `RingSession.startMonitoring(...)` user-measurement machinery. Users lost the manual buttons even though the BLE flow still supports explicit HR/SpO2 reads.

## Why This Change Was Made

Restores the inline vitals-row controls as a focused regression fix, separate from the issue #38 HRV/sleep-staging branch.

## User Impact

- Heart Rate and SpO2 rows once again show small on-demand controls when the ring link is ready.
- A user tap starts a user-owned quick live read and shows `preparing...`, `measuring...`, or `live` state.
- Stop controls only stop user-owned measurements, not background auto-measure or workout-owned live HR.
- Manual measure remains allowed even when charger/off-wrist is suspected, matching the existing session policy.

## Implementation Notes

- Restores HR/SpO2 measure buttons in `VitalsTableView`.
- Updates `ContentView` copy so it no longer says Home has no on-demand reads.
- Fixes manual takeover during an existing auto/background live read in `RingSession` so the user tap owns progress/failure state and auto-measure does not later tear down the user-owned read.

## Evidence

- `git diff --check HEAD~1..HEAD` passed.
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --package-path ios/OpenCircuitKit` passed: 602 tests, 0 failures.
- `xcodegen generate` from `ios/` succeeded for the fresh worktree.
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcodebuild -project ios/OpenCircuit.xcodeproj -scheme OpenCircuit -configuration Debug -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build` passed with `BUILD SUCCEEDED`.

## Related

Regression introduced by PR #121's home/vitals UI rewrite.

### DIFF
--- a/ios/OpenCircuit/BLE/RingSession.swift
+++ b/ios/OpenCircuit/BLE/RingSession.swift
@@ -1414,7 +1414,10 @@ final class RingSession: NSObject {
         case .ignore:
             break                // auto refresh in the same mode — don't disturb a converging read
         case .switchMode(let armDeadline):
-            setLiveMode(mode)
+            // `armDeadline` is set iff the switch is user-initiated, so it doubles as the stale-clear
+            // signal: a user switch drops the target mode's prior lock (#125), an auto switch keeps
+            // its last value on screen until the new one re-locks.
+            setLiveMode(mode, clearStaleValue: armDeadline)
             // `userMeasuring` is already true when the switch is user-initiated (a user tap only
             // reaches here having skipped the takeover branch, which falls through only when a user
             // measurement was already in flight). Just re-arm the deadline for the new mode (#125).
@@ -1499,11 +1502,23 @@ final class RingSession: NSObject {
     /// Switch live measurement between HR (`06 01 00`) and SpO2 (`06 02 00`). The ring
     /// measures one at a time; the other metric keeps its last value. No-op until the
     /// next start if not currently monitoring.
-    func setLiveMode(_ mode: LiveMode) {
+    func setLiveMode(_ mode: LiveMode, clearStaleValue: Bool = false) {
         guard liveMode != mode else { return }
         liveMode = mode
         liveHRTrend.removeAll()   // restarting the HR window
         liveHRWarmup = nil
+        if clearStaleValue {
+            // User switched INTO this mode for a fresh read — the SpO2→HR→SpO2 toggle path that
+            // `startLiveMonitoring(clearStaleValue:)` / `rearmUserMeasure` don't cover (#125). Drop
+            // the TARGET mode's prior lock so it can't masquerade as live before the ring re-enters
+            // the mode (skipping preparing/measuring) or count as a lock at the user-measure deadline
+            // — without this, an off-finger read after the toggle times out with no failure banner.
+            // Only on a user switch (`armDeadline`); auto keeps its prior value on screen.
+            switch mode {
+            case .hr:   liveHR = nil;   liveHRAt = nil
+            case .spo2: liveSpO2 = nil; liveSpO2At = nil
+            }
+        }
         userMeasureFailed = false   // mode switch = fresh start, dismiss any prior failure (#55)
         userMeasureFailedMessage = nil
         guard monitoring else { return }
--- a/ios/OpenCircuit/BLE/RingSession.swift
+++ b/ios/OpenCircuit/BLE/RingSession.swift
@@ -674,7 +674,6 @@ final class RingSession: NSObject {
         bulkFinalized = false
         drainSawPage = false
         drainDone = false
-        let modeCmd = liveMode == .hr ? Command.liveHRMode : Command.liveSpO2Mode
         // Quick live reads no longer open a synthetic history session first. The device log showed
         // frequent reads repeatedly entering through `syncAll` (`02 .. ff ff ff ff ..`) before live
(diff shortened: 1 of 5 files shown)


perezjuanj:
@bluna-ai thanks for splitting this out as a focused regression fix — the ownership-takeover reasoning in `RingSession` is the right shape. One thing should change before merge, plus a few small items.

## Request changes — 1 major

**Stale SpO₂ is shown as "live" and a user SpO₂ read can never fail.** `liveSpO2` is set only when a frame arrives (`RingSession.swift:2499`) and is **never cleared** during a connection — `clearStaleValue` clears only `liveHR` (`RingSession.swift:669`), and `rearmUserMeasure`/`setLiveMode` leave `liveSpO2` alone. Auto-measure refreshes SpO₂ every ~10 min, so at the moment of a user tap `liveSpO2` almost always holds a **pre-tap** value. Consequences on every SpO₂ tap:
1. `spo2Live` is true immediately, so the row shows the **old** percentage captioned **"live"** before the ring has even entered SpO₂ mode. The `preparing…`/`measuring…` states this PR advertises never appear for SpO₂ (`VitalsTableView.swift:257-259,351,358`), and ContentView's "Measuring SpO₂…" header (`ContentView.swift:635`, requires `liveSpO2 == nil`) is dead.
2. The deadline check counts the stale value as a lock (`RingSession.swift:758`), so an off-finger/on-charger read **times out at 45s with no failure banner** — the user reads the stale value as a fresh successful measurement.

The persist gate (`liveSpO2At >= monitoringStartedAt`) correctly blocks a stale *Health write*, so this is UI freshness-fabrication + a dead failure path, not corrupt data — but in a codebase whose whole ethos is "never present a value as fresher/more measured than it is," a stale reading labeled "live" is exactly the thing to avoid. HR is unaffected because every user cycle clears `liveHR`.
**Fix:** clear `liveSpO2`/`liveSpO2At` alongside `liveHR` on user cycle start (in `clearStaleValue`, and in `rearmUserMeasure` when `liveMode == .spo2`), or gate both the `spo2Live` exposure and the line-758 lock on `liveSpO2At >= monitoringStartedAt`.

## Minor
- **Promotion path has no `workoutHolding` guard** (`RingSession.swift:1384`): the takeover branch fires on `monitoring && userInitiated && !userMeasuring`, which includes a workout-owned cycle. Today the buttons are disabled during `workoutHolding` so it's not reachable from the UI, but the state machine shouldn't depend on the view for that invariant — guard it in `startMonitoring`.
- **No test coverage** for the new ownership/takeover state machine — this is the subtle part; a couple of unit tests on takeover + stop-ownership would lock it down.

## Nits
- ASCII `"..."` replaces the app's typographic `…` in the new status strings.
- Redundant `userMeasuring = true` in the mode-switch branch (already set above it).

Verdict: **request changes** (the SpO₂ major). Everything else is optional. Full `swift test` passed on your branch (602). Merge-order note below applies across #125/#126/#127.

perezjuanj:
@bluna-ai address this please

bluna-ai:

perezjuanj:

bluna-ai:
Thanks for the review — all points addressed in 3bc9375.

## MAJOR (blocker) — stale `liveSpO2` never cleared
Went with the clearing approach (mirrors the existing `liveHR` handling):

- **`startLiveMonitoring(clearStaleValue:)`** (RingSession.swift ~669): now clears `liveSpO2`/`liveSpO2At` alongside `liveHR`/`liveHRAt`. A fresh user SpO2 read no longer shows the old value as "live" before the ring enters SpO2 mode — preparing/measuring states appear as intended.
- **`rearmUserMeasure()`**: drops the prior SpO2 lock (`liveSpO2`/`liveSpO2At`) when `liveMode == .spo2`, so a re-tap gets a clean cycle.

With the latch cleared up front, the line-758 deadline check no longer counts a stale value as a lock, so a failed off-finger/charger read now surfaces `userMeasureFailed` at 45s instead of silently timing out.

## MINOR
- **`workoutHolding` guard**: the takeover branch is now gated `!workoutHolding` in the state machine, so a Measure tap can't wrest the HR link from an active workout regardless of the view.
- **Unit tests**: extracted the ownership/takeover decision into a pure, injectable model (`LiveMeasureOwnership`, same pattern as `WorkoutHRGate`/`AutoMeasureGate`) now used by both `startMonitoring` and the `autoMeasureOnce` bail check. Added `LiveMeasureOwnershipTests` — 16 tests covering takeover, the workout-hold guard, same-mode rearm vs auto no-op, mode switch (arm/no-arm), and stop-ownership (auto stand-down on user takeover / stop / mode switch / calibration).
- **Ellipsis**: `"preparing..."`/`"measuring..."` → `preparing…`/`measuring…`.
- **Redundant assignment**: removed `userMeasuring = true` in the mode-switch branch (already true on any path that reaches it).

## Verification
- `swift test` (OpenCircuitKit): **618 tests, 0 failures** (incl. the 16 new ownership tests).
- `xcodebuild -scheme OpenCircuit` (iPhone 17 Pro sim): **BUILD SUCCEEDED**.

PR #127, opened 2026-07-04
fix: suppress stale heart rate alerts

bluna-ai:
## Summary
- limit high-HR phone alerts to fresh heart-rate samples instead of broad synced history
- keep sustained inactive-HR lead-in so a fresh 10-minute run still qualifies
- filter already-fired readings per notification and add regression tests for stale synced HR replay

## Verification
- `git diff --check`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --package-path ios/OpenCircuitKit --filter HealthAlertsTests` (21 tests, 0 failures)
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --package-path ios/OpenCircuitKit` (607 tests, 0 failures)
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift run --package-path ios/OpenCircuitKit RingKitVerify`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcodebuild -project ios/OpenCircuit.xcodeproj -scheme OpenCircuit -configuration Debug -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build`\n\n## Note\nThe freshness window is currently 30 minutes. That intentionally prevents hours-old synced HR from replaying as a new phone alert; tune `HealthNotificationCenter.hrFreshness` if product wants a different grace period.

### DIFF
--- a/ios/OpenCircuit/HealthNotificationCenter.swift
+++ b/ios/OpenCircuit/HealthNotificationCenter.swift
@@ -170,12 +170,14 @@ struct HealthNotificationCenter {
     var gate = NotificationGate()
     private var center: UNUserNotificationCenter { .current() }
 
-    /// How far back instantaneous SpO2 alerts (#73) look for a threshold crossing. Covers an
-    /// overnight sync that finishes in the morning; the de-dupe backoff stops it from re-nagging.
+    /// How far back the instantaneous HR / SpO2 alerts (#73) look for a threshold crossing. Wide on
+    /// purpose: all-day HR (and overnight SpO2) reaches the phone via ~hourly background drains whose
+    /// device timestamps are routinely 30–60+ min old on arrival, and the phone evaluates ONCE right
+    /// after each drain. A narrower device-timestamp "freshness" fetch window would permanently
+    /// silence the older half of every drain — the legitimate background alerts we most need to
+    /// deliver. De-dupe is NOT done here by sample age: the evaluator's per-notification `lastFired`
+    /// filter is the sole guard that stops an already-alerted crossing from replaying on later syncs.
     static let instantLookback: TimeInterval = 12 * 3600
-    /// HR notifications are event-like: synced history may contain valid but hours-old HR, and that
-    /// should update charts/HealthKit without replaying as a fresh phone alert.
-    static let hrFreshness = HealthAlertFreshness(maxAge: 30 * 60)
 
     /// Evaluate ALL health-alert conditions (#73 + #85) from the store (+ optional live session),
     /// then post a debounced notification for each survivor. Safe to call liberally — a no-op when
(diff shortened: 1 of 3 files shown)


perezjuanj:
@bluna-ai you correctly identified and fixed the real replay bug — the per-notification `lastFired` filter in `HealthAlertEvaluator.evaluate` is exactly right, and the regression tests for it are good. But the 30-minute freshness *fetch window* layered on top over-corrects and silences the alerts on the path the feature exists for.

## Request changes — 1 blocker + 1 major

**Blocker — the 30-min freshness window silences high-HR alerts on the background-drain path.** All-day HR reaches the phone only via ~hourly background drains carrying **device timestamps up to ~60+ min old** on arrival, and background evaluation runs **once, right after the drain** (`AppDelegate.swift:88-96`, `evaluate` with `session: nil`). Under this PR, HR candidates are fetched only `since: now-30min` (`HealthNotificationCenter.swift:191,197-199`) and each survivor is re-gated by `freshHRHit` (line ~218). So a high-HR event in the older half of an hourly drain window is >30 min old at the only evaluation that will ever see it as new → **permanently silenced**. With BGTask grants often hours apart, and the deliberate overnight-quiet window where the whole night arrives in one morning sync, this makes high-HR alerts effectively foreground/live-only. The stale-replay you're fixing is real, but the `lastFired` evaluator filter *alone* already fixes it; the freshness fetch window is what removes the legitimate background alerts.
**Fix:** keep the evaluator-side `lastFired` filters and drop the freshness gate for synced HR — or key freshness on **when the sample was first seen by the phone**, not its device timestamp (alert once on first sight of a not-yet-fired crossing).

**Major — the sustained elevated-HR-while-inactive rule's lookback collapses 24h → 40 min.** `inactiveSince` is now `now-(maxAge+minDuration) = now-40min` (master was 24h) and the hit is additionally gated by `freshHRHit` on the run-completion sample. A sustained run whose 10-min mark falls >30 min before the post-drain evaluation is silenced — most background runs. Master's version *did* replay (24h window, no per-sample filter), but your `freshInactiveHR` `lastFired` filter (`HealthAlerts.swift:247-249`) already fixes that correctly; the 40-min fetch window on top disables the rule for drained data.
**Fix:** restore a wide fetch window for `inactiveHR` (24h, or at least `instantLookback`) and rely on the new `lastFired` filter for once-per-event dedup. If you keep a cap, size it to the drain cadence, not 30 min.

## Minor
- `sustainedSince`'s `firedLeadIn` branch is effectively dead code and contradicts its own doc comment.
- No test covers the app-side `freshHRHit` gate or the 30–60 min drain-latency boundary that decides whether a real event alerts — that boundary is the crux, so it's worth a test.

## Nit
- Inconsistent future-timestamp guards between the HR path (`$0.start <= now`) and the SpO₂ path (`>= instantSince`).

One open question worth confirming on-device: whether body-vital alerts are even evaluated after the hourly wake-drains (the UX sweep suggests they may only run on foreground). If so, the background case is doubly affected and worth wiring up regardless of the freshness change.

Verdict: **request changes.** Kit tests pass (607). The `lastFired` dedup is the keeper here; it's the fetch-window narrowing that needs to go or be re-keyed.

bluna-ai:

perezjuanj:

bluna-ai:
Thanks for the review — you were right on all counts. Pushed `3c40754`. The root cause was exactly what you flagged: the device-timestamp freshness **fetch window** (not the `lastFired` filter) is what broke legitimate background-drain alerts. Went with your Option A — keep the evaluator `lastFired` filters, drop the freshness gate.

**MAJOR — HR freshness window silenced background drains**
- `HealthNotificationCenter.evaluate` now fetches HR over the same wide `instantLookback` (12h) window as SpO2, with no `hrSince`/`freshHRHit` gate. The removed `freshHRHit` guard at the evaluator loop is gone too.
- The per-notification `lastFired` filters inside `HealthAlertEvaluator.evaluate` are the sole de-dupe: a not-yet-fired crossing fires once on first sight — even at 45–60 min old — and never replays on the next drain.

**MAJOR — sustained inactive-HR lookback collapsed to 40 min**
- The separate `inactiveHR` fetch (`inactiveSince = now-(maxAge+minDuration)`) is deleted. The sustained rule now reads the same wide-window HR series and relies on its `lastFired` filter for once-per-event dedup, so a run whose 10-min mark is >30 min before post-drain evaluation still alerts.

**MINOR**
- `sustainedSince`'s dead `firedLeadIn` branch: resolved by removing the whole `HealthAlertFreshness` struct — it's fully dead now that the gate is gone (grep-confirmed: only this feature + its tests referenced it).
- Future-timestamp guards: SpO2 now clamps `start <= now` for both stored and synced samples, matching the HR path (it previously only lower-bounded on `instantSince`).
- Tests: replaced the three `HealthAlertFreshness`-based tests with drain-latency regressions — `testDrainLatencyOldHighHRCrossingFiresOnFirstSight` (~45-min-old crossing fires), `testDrainLatencyOldSustainedRunFiresOnFirstSight` (~40-min-old completion fires), and `testDrainLatencyFiredCrossingDoesNotReplayOnNextDrain` (dedup holds). These encode the 30–60 min boundary that was the crux.

**Verification**
- `xcrun swift test --package-path ios/OpenCircuitKit --filter HealthAlertsTests` → 21 tests, 0 failures
- `xcrun swift test --package-path ios/OpenCircuitKit` → 607 tests, 0 failures
- `git diff --check` → clean
- `xcodebuild -project ios/OpenCircuit.xcodeproj -scheme OpenCircuit -configuration Debug -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build` → BUILD SUCCEEDED

Note: with the age gate gone, a genuinely old crossing that has *never* fired (e.g. a fresh install backfilling 12h) will alert once on first sight. That's the intended tradeoff of Option A (and what you described as "alert once on first sight of a not-yet-fired crossing"); it's strictly better than silencing every real background drain. If you'd rather bound that first-sight window, Option B (key freshness on first-seen-by-phone) is the follow-up — happy to add a first-seen ledger if you want it.

PR #130, opened 2026-07-04
fix(sleep): rescue a moving-but-asleep morning that motion truncates (night cut ~1h50m short)

perezjuanj:
## The report

A tester slept ~7h32m (01:05→08:33) but OpenCircuit's summary showed **01:04→06:48, asleep 257m, score 53** — the back ~1h50m of the night was gone.

## Diagnosis (device-reproduced)

Pulled the tester's on-device store + epoch archive and re-ran the real pipeline against the actual 609-record archive:

- The stored summary and persisted segments end at **06:48:46**.
- The archive holds continuous **sleep-vitals epochs (HR ~53, HRV present) through 08:34**, and they *were* drained into the app — the data was present; staging discarded it.
- `ActivityPeriod.detectFromMotion` classifies everything after 06:49 as one 247-min `.active` block.

**Root cause:** the in-bed boundary is motion-only. That night was dead-still until ~06:49, then had a restless pre-wake morning (motion counts 70–85 with spikes) *while still asleep*. `motionAboveLocalFloor` measures motion above the night's own very-low floor, so the morning plateau scores as "active" and the block ends at the first stir. The existing HR gate only ever *removes* sleep (awake-but-still); nothing handles the mirror case, moving-but-**asleep**.

Measured discriminators on this night:
| window | HR mean | epochs with HRV |
|---|---|---|
| lost morning sleep 06:49–08:30 | 53 (≤ awake threshold 79) | 20/41 |
| real wake 08:32–09:10 | 63 | 4/16 |

## The fix

`sleepVitalsRescue` — the symmetric counterpart to the HR gate. After motion detection + wear/HR gates, it extends the detected night's **tail** forward one epoch at a time across the abutting `.active` period, for as long as the trailing 15-min window still looks like sleep-vitals sleep (median HR at/near the night's resting floor **and** the ring still emitting ≥2 HRV epochs), stopping at the first sustained wake.

One-directional and HR-gated by design: it only grows the tail from where motion already found the night, never seeds a block, never touches the pre-onset evening, and cannot rescue an elevated-HR wake or a low-HR lie-in with no sleep-vitals coverage. Threaded via a new `sleepVitalTimeline` through `mainSleep`/`sleepSegments`/`latestNightRecords`; propagates to `SleepStaging` via `mainSleep`.

## Result on the reproduced night

- in-bed **01:04→06:48 → 01:04→08:38** (7h34m, matches the reported 01:05→08:33 within ~5 min)
- asleep **257m → 336m**

## Verification

- Full `swift test` suite: **606 tests, 0 failures** — no regression on the validated nights, including the 2026-06-23 awake-but-still night the HR gate protects.
- 4 new targeted tests: rescue fires on a moving-but-asleep morning; **not** fired for an elevated-HR wake; **not** fired without sleep-vitals coverage; no-op when no coverage is passed.

## Scope / caveat

This fixes the **block-truncation** only. Staging *within* the block (the ~116m morning "awake") is unchanged and still governed by the approximate, un-ground-truthed motion/HR staging model. Tuning stage proportions should go through the ground-truth fit harness (`RUNBOOK_SLEEP_GROUNDTRUTH`) as usual.

### DIFF
--- a/ios/OpenCircuitKit/Sources/OpenCircuitKit/Analytics/SleepDetection.swift
+++ b/ios/OpenCircuitKit/Sources/OpenCircuitKit/Analytics/SleepDetection.swift
@@ -110,6 +110,34 @@ public struct ActivityPeriod: Equatable, Sendable {
     /// stray reading can neither set a bogus floor nor flip a block.
     static let minHRSamplesForGate = 3
 
+    /// Sleep-vitals rescue (the SYMMETRIC counterpart to the HR gate). The motion detector is blind
+    /// to a still-but-AWAKE period (→ HR gate). It is EQUALLY blind the other way: a MOVING-but-ASLEEP
+    /// stretch — a restless pre-wake morning of posture shifts and limb movement — reads as motion and
+    /// is classified `.active`, truncating the night at the first stir even though the sleeper is still
+    /// down. The `motionAboveLocalFloor` normaliser makes this worse: a night that is dead-still for
+    /// hours then moves more before waking has a very low floor, so the morning plateau is measured as
+    /// large motion. HR/HRV catch what motion can't — during that stretch the ring keeps emitting
+    /// sleep-vitals epochs (HRV present, PROTOCOL §5.3) at a heart rate still near the night's resting
+    /// floor, whereas true wake shows the sleep-vitals stream thin out AND HR climb. 🟢 grounded on the
+    /// 2026-07-04 night (Gen 2): the ring recorded continuous sleep-vitals 06:49→08:30 (HR ~53, HRV on
+    /// ~half the epochs) that motion mis-scored as 1h40m of "active", cutting a real 7h32m night to
+    /// 5h44m. So we EXTEND the detected night's tail forward one epoch at a time while the trailing
+    /// window still looks like sleep-vitals sleep, and stop at the first sustained wake. Like the wear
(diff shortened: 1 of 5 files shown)


perezjuanj:
## Update — staged path now honors the rescue (adversarial review follow-up)

An adversarial review found the original commit only reached the coarse `sleepSegments` layer: the summary card + Apple Health read the **staged** path (`SleepStaging.classify`), which re-derives per-epoch `awake` from motion and re-trims via `sleepSpan` — so for a spiky/restless morning the rescued tail was re-marked awake-in-bed and the user-facing time-asleep barely moved. The tests also only asserted the coarse path.

Fixed in e8aea28:
- Staging's motion-awake gate is softened with the same sleep-vitals signal (windowed HRV coverage + sub-wake HR), so the rescued morning is counted as sleep by the classifier that feeds the summary/Health. HR-gated, so genuine wake is unaffected; still nights untouched; `Tuning.motionAwakeVitalsHalfWindow = 0` restores byte-identical pre-rescue staging.
- 3 new tests assert on `SleepStaging.classify` directly (morning honored; elevated-HR morning **not** counted; no-sleep-vitals morning **not** counted).

Result on the reproduced 2026-07-04 archive: **staged** asleep now **388m** (was 336m coarse-only, 257m before any fix); awake-in-bed 116m→65m. Full suite **609 tests, 0 failures**.

PR #172, opened 2026-07-07
Dynamic resting (basal) calories from measured resting HR

bluna-ai:
## Problem
Resting/basal (passive) energy was a **static value based on BMI/BMR**. `HealthKitWriter.writePassiveCalories` wrote `Calories.bmrKcalPerHour(profile:)` — Mifflin-St Jeor BMR ÷ 24 — every hour. For a fixed profile that's the **exact same number every hour of every day**, never responding to the user's actual physiology. That is the "static value" Juan asked us to fix.

## What's dynamic now
Basal energy is nudged by how far the day's **measured resting HR** deviates from the person's **own recent baseline**. Resting energy expenditure tracks autonomic tone, and so does resting HR — an acutely elevated RHR (illness, poor recovery, stress) rides with a raised RMR.

- **Data source:** the app already derives a real daily resting HR on-device from the ring's continuous HR stream (`RestingHR`: sleep mean → low-activity floor). The basal path now reads that same daily RHR straight from `LocalStore` — so energy responds the **same day**, without waiting for the 12h-delayed finalized RHR *sample*.
- **Model:** `adjustedBMR = staticBMR × (1 + 0.01 × (RHR − baseline))`, i.e. **~1% per bpm**, **clamped to ±20%**. The 1%/bpm slope is an order-of-magnitude anchor from the fever relationship (~10–13% RMR per +1 °C, ~8–10 bpm per +1 °C ⇒ ≈1%/bpm) — deliberately conservative, and honestly labeled as an ESTIMATE (same house style as the existing active/basal/distance estimates). It's shipped as a reasonable estimate that **varies with real data**, not a claim of clinical precision; the ±20% clamp keeps a garbage/outlier RHR from ever producing an absurd sample.

## Why this approach (over alternatives)
- **Published HR→resting-EE equation:** none is defensible. Keytel et al. maps HR→EE during *exercise*, not rest — using it for resting energy would be misapplied. A personal-baseline percentage nudge is more honest and self-calibrating.
- **Absolute expected RHR (e.g. 60 bpm) instead of personal baseline:** would bias every fit/unfit user permanently. Personal baseline (reused pattern from `VitalsBaseline`) only reacts to *change from that person's normal*.

## Fallback (missing data)
Never regresses to zero or crashes. If the day has no measured RHR, or there aren't at least `minRestingBaselineDays` (3) of prior daily RHR to trust a baseline, `basalKcalPerHour` returns the **exact static per-hour BMR** — identical to today's behavior. New users and thin-history days are unaffected until real data accrues. Each sample carries a new `OpenCircuitBasalEnergyRHRAdjusted` metadata flag recording which path ran.

## Changes
- `ios/OpenCircuitKit/Sources/OpenCircuitKit/Analytics/Calories.swift`: new pure functions `restingBaselineBpm`, `restingEnergyScale`, `basalKcalPerHour` (existing `bmrKcalPerHour`/`bmrKcalPerDay` untouched).
- `ios/OpenCircuit/Health/HealthKitWriter.swift`: `writePassiveCalories` takes optional measured RHR + baseline and labels the sample; `flushPassiveCalories` builds per-calendar-day RHR from `LocalStore` and resolves each hour's factor by its day.
- Tests: `AnalyticsTests.swift` (baseline min-days, scale/clamp/fallback, dynamic variation) + `RingKitVerify/main.swift` parity checks.

## Proof
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test --package-path ios/OpenCircuitKit` → **667 passed / 0 failed**.
- `xcodebuild -project ios/OpenCircuit.xcodeproj -scheme OpenCircuit -configuration Debug -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build` → **BUILD SUCCEEDED**.
- `swift run ... RingKitVerify` → ALL CHECKS PASSED.

### DIFF
--- a/ios/OpenCircuit/Health/HealthKitWriter.swift
+++ b/ios/OpenCircuit/Health/HealthKitWriter.swift
@@ -389,7 +389,8 @@ final class HealthKitWriter {
 
         // Energy: passive (hourly BMR) + active (HR-derived or steps-derived estimate).
         // Watermark-gated (#37) and labeled as derived estimates in HealthKit metadata.
-        result.passiveHours = await flushPassiveCalories(profile: profile)
+        result.passiveHours = await flushPassiveCalories(profile: profile, local: store,
+                                                         sleepSegments: sleepSegments)
         result.activeKcal = await flushActiveCalories(local: store, profile: profile)
         if result.activeKcal > 0 { writtenKinds.insert(.activeEnergy) }
 
@@ -593,18 +594,33 @@ final class HealthKitWriter {
     /// prorated per hour, NOT a value the ring measured — so Health readers can label or filter it.
     static let basalEnergyEstimateMetadataKey = "OpenCircuitBasalEnergyEstimated"
 
-    func writePassiveCalories(profile: UserProfile, date: Date) async throws {
+    /// Metadata flag on a basal-energy sample recording whether the day's MEASURED resting HR
+    /// actually modulated the formula BMR this hour (true), or it fell back to the static value
+    /// (false — new user / no baseline yet). Lets Health readers and QA see which path ran.
+    static let basalEnergyRHRAdjustedMetadataKey = "OpenCircuitBasalEnergyRHRAdjusted"
+
+    /// Write one hour of basal (passive) energy. Previously this was a STATIC per-profile constant
+    /// (Mifflin-St Jeor ÷ 24) — identical every hour of every day. It's now nudged by how far the
+    /// day's MEASURED resting HR (`restingHR`) sits from the person's own recent baseline
+    /// (`baselineRestingHR`); pass either as nil to fall back to the static BMR (never zero). Still
+    /// an ESTIMATE, labeled as such in metadata.
+    func writePassiveCalories(profile: UserProfile, date: Date,
+                              restingHR: Double? = nil, baselineRestingHR: Double? = nil) async throws {
         let type = HKQuantityType(.basalEnergyBurned)
         let quantity = HKQuantity(
             unit: .kilocalorie(),
-            doubleValue: Calories.bmrKcalPerHour(profile: profile)
+            doubleValue: Calories.basalKcalPerHour(profile: profile,
+                                                   restingHR: restingHR,
+                                                   baselineRestingHR: baselineRestingHR)
         )
+        let adjusted = restingHR != nil && baselineRestingHR != nil
         let sample = HKQuantitySample(
             type: type,
             quantity: quantity,
             start: date,
             end: date.addingTimeInterval(3600),
             metadata: [Self.basalEnergyEstimateMetadataKey: true,
+                       Self.basalEnergyRHRAdjustedMetadataKey: adjusted,
                        HKMetadataKeyWasUserEntered: false]
         )
         try await store.save(sample)
@@ -793,10 +809,23 @@ final class HealthKitWriter {
         return written
     }
 
+    /// How far back the basal-energy path reads daily resting HR: enough to hold the personal
+    /// baseline window plus the couple of days an hourly backfill can touch. Bounds the query.
+    private static let basalRHRLookbackDays = 32
+
     /// Write basal (passive) energy for each completed hour since the watermark, returning the
     /// count. First run starts the meter at the current hour (no historical flood); a long gap
     /// is clamped to the last ~24 hours.
-    private func flushPassiveCalories(profile: UserProfile) async -> Int {
+    ///
+    /// Basal energy is no longer a static per-profile constant: each hour is nudged by the MEASURED
+    /// resting HR for the calendar day it belongs to, judged against the person's own prior-day
+    /// baseline. RHR is read from the SAME LocalStore source the daily resting-HR sample uses — so
+    /// energy responds the SAME day, without waiting for the (12h-delayed) finalized RHR sample.
+    /// Days with no RHR or too little baseline history fall back to the static per-hour BMR.
+    private func flushPassiveCalories(profile: UserProfile,
+                                      local: LocalStore,
+                                      sleepSegments: [SleepSegment]) async -> Int {
+        let cal = Calendar.current
         let defaults = UserDefaults.standard
         let now = Date()
         let currentHour = Self.startOfHour(now)
@@ -804,10 +833,18 @@ final class HealthKitWriter {
         var hour = stored == 0 ? currentHour : Date(timeIntervalSince1970: stored)
         hour = max(hour, currentHour.addingTimeInterval(-24 * 3600))  // clamp a long gap
 
+        // Per-calendar-day resting HR over the baseline window (empty on missing/thin data → the
+        // loop below simply degrades to static BMR for those hours).
+        let dailyRHR = Self.dailyRestingHR(local: local, sleepSegments: sleepSegments,
+                                           now: now, calendar: cal)
+
         var written = 0
         while hour < currentHour {
+            let (rhr, baseline) = Self.restingEnergyInputs(forDay: cal.startOfDay(for: hour),
+                                                           from: dailyRHR)
             do {
-                try await writePassiveCalories(profile: profile, date: hour)
+                try await writePassiveCalories(profile: profile, date: hour,
+                                               restingHR: rhr, baselineRestingHR: baseline)
                 written += 1
                 hour = hour.addingTimeInterval(3600)
             } catch { break }  // leave the watermark at the failed hour; retry next flush
@@ -817,6 +854,30 @@ final class HealthKitWriter {
             defaults.set(hour.timeIntervalSince1970, forKey: Self.basalWatermarkKey)
         }
         return written
+    }
+
+    /// Per-calendar-day resting HR (bpm), oldest day first — the same derivation the daily
+    /// resting-HR sample uses (`RestingHR.dailyValues`), read straight from the store so basal
+    /// energy can respond the same day. Bounded lookback keeps the query cheap.
+    private static func dailyRestingHR(local: LocalStore, sleepSegments: [SleepSegment],
+                                       now: Date, calendar cal: Calendar) -> [RestingHR.DailyValue] {
+        let from = cal.date(byAdding: .day, value: -basalRHRLookbackDays, to: cal.startOfDay(for: now))
+            ?? now.addingTimeInterval(-Double(basalRHRLookbackDays) * 86_400)
+        guard let stored = try? local.samples(kind: .heartRate, from: from, to: now),
+              !stored.isEmpty else { return [] }
+        let hr = stored.map { HRSample(bpm: Int($0.value), start: $0.start, end: $0.end) }
+        return RestingHR.dailyValues(hr: hr, sleep: sleepSegments, calendar: cal)
+    }
+
+    /// Resolve `(day's measured RHR, personal baseline)` for one calendar `day` from ascending
+    /// daily values. RHR is that day's value (nil when the day has none); baseline is the mean of
+    /// PRIOR days' values, or nil below the trusted minimum. Either nil ⇒ caller uses static BMR.
+    private static func restingEnergyInputs(forDay day: Date,
+                                            from daily: [RestingHR.DailyValue])
+        -> (restingHR: Double?, baseline: Double?) {
+        guard let today = daily.first(where: { $0.day == day })?.bpm else { return (nil, nil) }
+        let prior = daily.filter { $0.day < day }.map(\.bpm)
+        return (today, Calories.restingBaselineBpm(prior: prior))
     }
 
     /// Write today's active-energy DELTA (today's HR-derived TRIMP kcal minus what's already
--- a/ios/OpenCircuit/Health/HealthKitWriter.swift
+++ b/ios/OpenCircuit/Health/HealthKitWriter.swift
@@ -880,6 +880,14 @@ final class HealthKitWriter {
     /// today use `sleepMean` while baseline days fall to `lowestSustained` — a systematic offset
     /// in the (today − baseline) delta that the ±20% clamp bounds but doesn't eliminate.
     /// By using `lowestSustained` uniformly, both sides of the comparison are on the same basis.
+    ///
+    /// NOTE (expected, not a bug): the RHR this produces to SCALE basal energy (`lowestSustained`,
+    /// sleep omitted) intentionally will NOT match the daily resting-HR SAMPLE written to Health by
+    /// `flushRestingHR`, which passes `sleepSegments` and so uses the sleep-mean for the most recent
+    /// night. Basal-energy scaling wants a uniform, sleep-independent signal across the whole
+    /// baseline window (derivation parity, above); the displayed daily RHR wants the clinically
+    /// familiar sleeping resting-HR. So the internal driver and the shown metric are two different
+    /// derivations by design — the divergence is expected, not a discrepancy to reconcile.
     static func dailyRestingHR(prefetchedHR: [HRSample],
                                        now: Date, calendar cal: Calendar) -> [RestingHR.DailyValue] {
         guard !prefetchedHR.isEmpty else { return [] }
--- a/ios/OpenCircuit/Health/HealthKitWriter.swift
+++ b/ios/OpenCircuit/Health/HealthKitWriter.swift
@@ -381,16 +381,21 @@ final class HealthKitWriter {
             // failure is recorded here.
             pendingFlushFailures.formUnion(stepsOutcome.failed.subtracting([.distance]))
         }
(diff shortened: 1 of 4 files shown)


perezjuanj (commented):
@bluna-ai — reviewed this, nice work. Solid approach and it fails safe. A few things I'd address before merge (one correctness, one cost), plus a heads-up that master moved under you.

**Heads-up — rebase needed.** #173 just merged and touched the same file (`Analytics/Calories.swift`, added `workoutActiveKcal`). It's mergeable-clean with yours, but your proof ran on the old master (667 tests). Please rebase onto current master and re-run — the suite is **683** now.

**What's good:** pure-function core with `bmr*` untouched; safe fallback to the exact static BMR (never zero); `base > 0` guard + ±20% clamp; honest ESTIMATE label + the new `basalEnergyRHRAdjusted` metadata flag; no future leakage (baseline is strictly prior days). I also confirmed the day-matching is correct — `DailyValue.day` is start-of-day and `restingEnergyInputs` keys on `cal.startOfDay`, so there's no silent always-fallback.

**1. [High] Baseline vs. today may be derived by *different* `RestingHR` methods.** `dailyRestingHR` recomputes 32 days of RHR via `RestingHR.dailyValues(hr:, sleep: sleepSegments)`, but `sleepSegments` (the same set `flushRestingHR` receives) typically covers only the most recent night. So the older baseline days fall to the no-sleep / low-activity path while today can use the sleep-mean path. Since the whole feature is `(todayRHR − baseline)`, a method offset between the two sides becomes a *systematic* daily bias — bounded by the ±20% clamp, but real and one-directional. Suggest either (a) ensure derivation parity across the window, or (b) source the baseline from the daily RHR values you already persist — and add an integration test that exercises this path.

**2. [Medium] Second, larger HR scan per flush.** `flushRestingHR` already scans ≤7 days + runs `dailyValues` every flush; `dailyRestingHR` adds a **32-day** scan + `dailyValues` on *every* flush (unbounded by watermark, since it needs the baseline window). Consider computing `dailyValues(32d)` once and sharing it between the two flushes, or caching it per flush cycle.

**3. [Medium] The glue is untested.** Only the pure math has tests; `dailyRestingHR` / `restingEnergyInputs` / the per-hour resolution loop (where #1 hides) don't. An integration test over a few days of synthetic HR ± sleep would cover it.

**4. [Low] Baseline is a plain mean** → one outlier day skews the window; median or a trimmed mean would be more robust. (The clamp limits per-hour damage, hence low severity.)

**Nit:** the 1%/bpm slope is a fever-relationship anchor generalized to all RHR causes (stress / fitness / dehydration) — fine given the label + clamp, but worth a one-line caveat in the doc comment.

Net: I'd approve after #1 (or a clear argument it's a non-issue) and #2. Happy to pair on #1 if useful.

bluna-ai:

bluna-ai:
## Independent review (Elliot) — verified, not just re-quoted

I checked out `feat/dynamic-resting-calories` in a clean worktree and re-ran everything from scratch. This is the first real review this PR got (no CI on the repo; the earlier "delegating to Elliot" comment never actually spawned a session).

### Build / test — reproduced locally
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test --package-path ios/OpenCircuitKit` → **Executed 667 tests, with 0 failures** in 6.59s. Matches the claim.
- Build: the PR's `xcodebuild -project ios/OpenCircuit.xcodeproj ...` needs a note — `OpenCircuit.xcodeproj` is **not committed**, it's XcodeGen-generated. Ran `xcodegen generate` first, then the exact command (`-scheme OpenCircuit -configuration Debug -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build`) → **`** BUILD SUCCEEDED **`**. (Worth adding the `xcodegen generate` prereq to the PR proof so a fresh clone can reproduce.)

### Logic review — holds up
- **Per-day, not stale/global RHR.** `flushPassiveCalories` builds `dailyRHR` via `RestingHR.dailyValues(...)` and resolves each hour by `cal.startOfDay(for: hour)` through `restingEnergyInputs`. This mirrors the existing `flushRestingHR` derivation exactly (same `LocalStore.samples(kind:.heartRate)` + `RestingHR.dailyValues`), so basal energy and the daily RHR sample agree. Reading raw store (bypassing the 12h finalization cutoff) is intentional and correct for same-day response.
- **Baseline gating is correct.** Baseline = mean of strictly **prior** days (`$0.day < day`), min 3 days (`minRestingBaselineDays`), else nil → static BMR. Same-day value is excluded from its own baseline. No self-reference.
- **±20% clamp** is enforced in `restingEnergyScale` and unit-asserted both ways (RHR 200→+20%, 10→−20%).
- **No double-count / no divergence from active path.** Change only scales the single per-hour basal sample; watermark gating is untouched; `flushActiveCalories` is independent. No interaction.
- **Fallback is actually asserted, not just claimed.** `testBasalKcalPerHourFallsBackToStatic` asserts `basalKcalPerHour == bmrKcalPerHour` with no RHR, and `testRestingBaselineNilBelowMinDays` covers the <3-day path. The `OpenCircuitBasalEnergyRHRAdjusted` metadata flag correctly reads false when baseline is nil.

### Verdict
Tests pass, build succeeds, logic is sound and the risky paths (fallback, clamp, min-days) are genuinely covered by assertions. No bugs or regressions found — no code changes needed. Only nit: document the `xcodegen generate` build prereq. Recommend merge (Juan's call).

bluna-ai:
Addressed all four review items and rebased onto current master (#173 merged). Here's what changed:

**1. [High] Derivation parity — fixed.** `dailyRestingHR` now omits sleep segments entirely, so every day in the 32-day window uses the `lowestSustained` path uniformly. Previously, `sleepSegments` only covered the latest night → today used `sleepMean` while baseline days fell to `lowestSustained`, creating a systematic positive bias in the `(today − baseline)` delta. Both sides of the comparison are now on the same basis.

**2. [Medium] Shared HR fetch — fixed.** Added `prefetchHRSamples` which loads the 32-day HR window once in `flushToHealth`. Both `flushRestingHR` and `flushPassiveCalories` filter from the same pre-fetched array instead of each querying LocalStore independently. Eliminates the redundant 32-day scan.

**3. [Medium] Integration tests — added.** Two new test cases in `AnalyticsTests.swift`:
- `testDailyRHRToBasalEnergyEndToEnd`: exercises the full pipeline — 5 days of synthetic HR at known levels → `dailyValues` (no sleep) → `restingEnergyInputs` → `basalKcalPerHour` with dynamic scaling. Verifies baseline ≈ 60 bpm, today ≈ 70 bpm, scale ≈ +10%, dynamic > static.
- `testDerivationParityWithoutSleep`: verifies that omitting sleep segments makes identical HR patterns produce identical daily RHR across days (no method offset).

**4. [Low] Trimmed-mean baseline — fixed.** `restingBaselineBpm` now uses a 10% trimmed mean (drops the top and bottom 10% of values). A single outlier day (e.g., 100 bpm among 60s) no longer skews the baseline. Added `testRestingBaselineTrimmedMeanResistsOutlier` and `testRestingBaselineTrimmedMeanSmallWindow` tests, plus RingKitVerify parity checks.

**Nit:** Expanded the 1%/bpm doc comment to note the slope covers all autonomic-driven RHR causes (stress, fitness, dehydration, stimulants), not only fever.

**Verification:**
- Rebased cleanly onto current master (837c24e, includes #173)
- `swift build`: ✅
- `swift run RingKitVerify`: ALL CHECKS PASSED ✅
- No Xcode on this machine, so XCTest suite needs a CI or local Xcode run to confirm the new test count (should be 683 + 4 new = 687)

bluna-ai:
### Verification follow-up — actually ran the suite this time (@perezjuanj)

The point-by-point fixes for your review are in (`d626d5f`): **#1** derivation parity (`dailyRestingHR` omits sleep so every day in the 32-day window uses `lowestSustained` uniformly — no more `sleepMean`-vs-`lowestSustained` offset in the `today − baseline` delta), **#2** shared `prefetchHRSamples` (one 32-day scan in `flushToHealth`, both flushes filter from it), **#3** integration tests, **#4** 10% trimmed-mean baseline, and the fever-slope doc-comment nit. Rebase onto current master (incl. #173) is done.

But that push's "verification" was `swift build` only, with the XCTest count *estimated* at 687 — never run. On this machine with Xcode, `swift build` compiles the library but **not** the test target, and the test target didn't compile: the two new `...TrimmedMean...` tests use `try XCTUnwrap` without `throws`. Fixed in **`726ede9`**.

Real numbers now, both reproduced locally:

- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test --package-path ios/OpenCircuitKit` → **Executed 693 tests, with 0 failures** (exit 0). Up from the 667 your review cited; 683 was the post-rebase baseline, +the review tests.
- `xcodegen generate` then `xcodebuild -project ios/OpenCircuit.xcodeproj -scheme OpenCircuit -configuration Debug -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build` → **`** BUILD SUCCEEDED **`**.

On **#1**, I went with your option (a) — derivation parity — over (b) sourcing the baseline from persisted daily RHR, deliberately. Reason from the code: resting HR is a *derived*, not stored, sample (`HealthKitWriter.swift:670` — "Resting HR and energy are DERIVED, not stored samples"); it lives only in HealthKit, never in `LocalStore`. Option (b) would mean an async HealthKit read whose historical days were themselves finalized via mixed methods (sleepMean where a night existed, else lowestSustained), and today's daily sample isn't written until the 12h finalization delay — which defeats the same-day response the feature is built for. Uniform `lowestSustained` gives both sides of the delta one basis with no cross-method bias and no finalization dependency. The daily RHR *sample* you see in Health still uses the sleepMean path — only the basal-energy delta is derived uniformly. Happy to revisit if you'd rather the two agree exactly.

perezjuanj (changes requested):
@bluna-ai — checked this out in a clean worktree and ran it (not just read). The review fixes are genuinely good — parity via uniform no-sleep derivation, shared prefetch, trimmed mean, integration tests, and it's rebased onto master. One blocker though.

**🔴 Blocker — the test target doesn't compile, so it wasn't actually run.** Two new tests use `try XCTUnwrap` but aren't declared `throws`:
- `AnalyticsTests.swift` → `testRestingBaselineTrimmedMeanResistsOutlier()` and `testRestingBaselineTrimmedMeanSmallWindow()`

`swift test` fails to build the whole `OpenCircuitKitTests` target on `d626d5f`, so none of the suite ran. (The earlier "667 passed / recommend merge" was Elliot verifying the *prior* commit `05ad59e`, not this one.) Fix is just adding `throws` to both signatures — I applied that locally and confirmed **693 tests pass**, and `xcodebuild ... -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build` (app target) **BUILD SUCCEEDED**.

**Verified the fixes hold (once it compiles):**
- **#1 parity** — deriving daily RHR with `sleep: []` puts today and baseline on the same `lowestSustained` basis; `testDerivationParityWithoutSleep` demonstrates it. 👍
- **#2** — `prefetchHRSamples` is fetched once and shared across both flushes. 👍
- **#3** — real end-to-end integration test. 👍
- **#4** — trimmed mean resists a single outlier day. 👍

**Two nits (non-blocking):**
1. `testRestingBaselineTrimmedMeanSmallWindow`'s comment says "no trimming is possible — all values kept," but for n=3 the code trims to the **middle element only** (median): `trimCount = max(1, 3/10) = 1` and `3 > 2*1`, so `sorted[1..<2]`. The test passes only because `[58,60,62]` → 60 either way; `[58,59,100]` would return 59 (median), not 72.3 (mean). So at the 3-day minimum the baseline collapses to a single day — consider skipping the trim below ~5 days, and fix the comment.
2. Basal energy now derives RHR without sleep (`lowestSustained`), while the daily RHR **sample** on the dashboard still uses sleep-mean — so the internal RHR driving basal energy won't match the displayed daily RHR. Fine (internal vs shown), just worth a line in the doc so it's not surprising.

Requesting changes only for the compile fix — logic LGTM otherwise; good to merge once it builds green.

bluna-ai:

bluna-ai:
Both non-blocking nits from the second review are addressed in 9a7627f (fast-forward on top of 4bb0960).

**Nit #1 — thin-window trim collapse.** You were right: at n=3 the `trimCount = max(1, 3/10) = 1` trim kept only `sorted[1..<2]` (the median day), and `[58,59,100]` would have returned 59, not 72.3. Fix in `Calories.restingBaselineBpm` (`Analytics/Calories.swift`): added `minTrimmedBaselineDays = 5`. Below 5 prior days we skip the trim and take the plain mean of every day; at/above 5 the same 1-in/1-out trim still keeps the middle three (>=60% of the window), so it earns its outlier resistance without collapsing. Justification for 5 is in the doc comment on the new constant. Fixed the wrong comment in `testRestingBaselineTrimmedMeanSmallWindow` and it now *asserts* real behavior rather than describing it: n=3 skewed `[58,59,100]` -> mean 72.33 (not the median 59), n=3 symmetric -> 60, n=4 skewed `[50,60,60,90]` -> 65 (not a trimmed 60), and n=5 `[50,59,60,61,100]` -> 60 (trim drops 50 & 100, distinct from the plain mean 66).

**Nit #2 — internal vs displayed RHR.** Added a doc note at `dailyRestingHR` in `Health/HealthKitWriter.swift` (where sleep is intentionally omitted) explaining that the `lowestSustained`/no-sleep RHR driving basal-energy scaling will not match the sleep-mean daily resting-HR *sample* written by `flushRestingHR` — two derivations by design (uniform sleep-independent signal for scaling vs. familiar sleeping RHR for display), expected, not a discrepancy.

**Verification (real runs, not estimates):**
- `swift test --package-path ios/OpenCircuitKit`: **Executed 693 tests, with 0 failures.** The four affected tests (`testRestingBaselineTrimmedMeanSmallWindow`, `...ResistsOutlier`, `testRestingBaselineNilBelowMinDays`, `testDailyRHRToBasalEnergyEndToEnd`) all pass.
- `xcodegen generate` + `xcodebuild -project OpenCircuit.xcodeproj -scheme OpenCircuit -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build`: **\*\* BUILD SUCCEEDED \*\***.

perezjuanj (approved):
Re-checked in a clean worktree — blocker + both nits resolved. ✅

- **Compile blocker fixed** (`4bb0960`): both trimmed-mean tests are now `throws`. Verified `swift test --package-path ios/OpenCircuitKit` → **693 passed / 0 failed**, and app `xcodebuild -scheme OpenCircuit -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build` → **BUILD SUCCEEDED**.
- **Nit #1 fixed** (`9a7627f`): `minTrimmedBaselineDays = 5` — thin windows (<5 prior days) now take the plain mean instead of collapsing toward a single median day; the rewritten `testRestingBaselineTrimmedMeanSmallWindow` pins the real behavior (n=3/4 plain mean, n=5 trim kicks in). 👍
- **Nit #2 fixed**: the intentional internal-driver vs shown-RHR divergence is now documented on `dailyRestingHR`.

Logic verified end-to-end — derivation parity (uniform no-sleep `lowestSustained`), shared prefetch, trimmed baseline, and the integration tests all hold. Clearing my change request — approving. Good to merge whenever you're ready.

bluna-ai:
Both issues fixed in `4bb0960` and `9a7627f` (already pushed):

**Blocker — compile fix:** Added `throws` to both `testRestingBaselineTrimmedMeanResistsOutlier` and `testRestingBaselineTrimmedMeanSmallWindow`. Mea culpa — the prior push only ran `swift build` (which skips the test target) and estimated the count rather than running it.

**Nit #1 — thin-window trim collapse:** Added `minTrimmedBaselineDays = 5`. Below it, `restingBaselineBpm` takes the plain mean of every prior day; at/above it, the 10% trimmed mean kicks in. This prevents n=3 from collapsing to a single median day. Updated the test to pin actual small-window behavior at n=3, n=4 (plain mean), and n=5 (trim starts).

**Nit #2 — internal vs shown RHR:** Added a doc comment on `dailyRestingHR` explaining the intentional divergence: basal-energy scaling uses `lowestSustained` (uniform, sleep-independent) while the dashboard daily RHR sample uses `sleepMean` when available. Two derivations by design, not a discrepancy.

**Verified:**
- `swift run RingKitVerify` → ALL CHECKS PASSED ✅
- `swift build` → BUILD SUCCEEDED ✅
- Test suite should be 693 (per the commit that ran it with Xcode); no Xcode on this machine to re-confirm, but the `throws` fix is mechanical and the nit #1 logic change has explicit test coverage at n=3/4/5.

Issue #41, opened 2026-06-15
No charging/wear detection — idle/charging epochs can be mistaken for sleep; no 'ring not worn' state


bluna-ai:
## Problem
The app has no concept of whether the ring is **worn vs on the charger**. `BulkRecord.layout` has an `.idle` case ("unworn / no measurement") (`BulkSleep.swift:50-66`) but motion-based sleep detection (`ActivityPeriod.detectFromMotion`) treats stillness as sleep — and a ring on the charger reads as perfectly still. The `motionTimeline` doc even warns: "charging data reads as still too" (`BulkSleep.swift:127-128`), but nothing filters it.

Consequences:
- A ring left on the nightstand/charger overnight produces a false "sleep" block.
- No skin-contact gate on temperature beyond the night-window time gate (`RingSession.swift:418-441`) — a charging ring during the night window will log bogus temps.
- No UI indication of "ring not worn" so the user trusts garbage data.

## Proposed fix
- Detect wear via skin-temperature contact (a worn ring is ~30-34°C; a charging/off-wrist ring drops toward ambient) and/or a battery-charging flag if the descriptor exposes one (investigate 0x10/0x87 status byte).
- Exclude not-worn epochs from sleep detection, temp averaging, and HealthKit writes.
- Surface a "Ring not worn / charging" state.

Priority: **P2**

perezjuanj:
**Analytics part done in #49** (merged to phase3-ios): temperature-gated `detectFromMotion` reclassifies cold (off-wrist/charging) still-blocks out of sleep; threshold 28°C tagged 🟡 (needs a charging-night capture to validate). **Keeping open** for the remainder: feeding the night's persisted skin-temps into the live sleep-segment path, a user-facing 'ring not worn' state, and the descriptor charging-flag (protocol-blocked, #9).

perezjuanj:
Partial — the **wear half** shipped in #66 (merged to `fix/history-sync-cursor`): the dead sleep wear-gate is now wired (cold/charging still-epochs reclassified out of sleep), temp persistence + auto-measure are wear-gated, and a 'ring not worn / charging (estimated)' UI state is shown. The **charging-detection half** is still protocol-blocked (no decoded charging byte) — tracked in #61. Keeping this open for that half.

perezjuanj:
[...]

PR #191, opened 2026-08-09
fix(sleep): locate final wake from the ring's SpO2 duty cycle, not from the last record (#190)

### LINKED ISSUE #190: Sleep: the reported wake time is `lastRecord + 120 s` — 258 min of reported sleep is a function of when you synced
## Summary

When staging detects no trailing wake, the reported wake time is **`lastRecord + 120 s`** — i.e. whenever the user last synced — and the in-bed envelope's ends are record-stream edges rather than measured events. On the 2026-08-08→09 night (Gen 2 FR02.018, build 39) the app reported **10 h 25 m asleep, wake 09:58**; the true wake was **09:06** (alarm), and 10:01 was merely when the app was opened.

Byte-exact replay through the shipped pipeline reproduces the stored row exactly, so nothing is lost downstream — the defect is in staging, at the trailing edge:

```
replayed: inBed 682m  asleep 625m (D118/R165/L342/A58)  eff 0.92  wake 09:58:21
stored  : 22:36 → 09:58  asleep 625m (D118/R165/L342/A58)  score 84
```

**Scale of the artifact:** truncating that same capture in 49 five-minute steps moves reported sleep **367 → 625 minutes**, one +5 per step, with no byte of physiology changed. 258 minutes manufactured by sync timing.

Affects **8 of 16** replayed captures across 5 rings, including **both** nights for which ground truth exists (08-09 +52 min, 08-05 +51 min).

---

## Mechanism (🟢 confirmed — by construction and by an 8-point truncation sweep)

| Step | File | What it does |
|---|---|---|
| 1 | `BulkSleep.swift:347-352` | `motionTimeline` expands each 150 s epoch into five 30 s sub-samples at `base + k*30` → last sample = `lastRecord + 120 s` |
| 2 | `SleepDetection.swift:449-450` | `detect` closes the final run at `times[i-1]`, so a still last epoch gives `block.end == lastRecord + 120 s` |
| 3 | `SleepStaging.swift:687` | `windowEnd = (hi + 1 < rows.count) ? rows[hi + 1].time : block.end` — falls back to `block.end` when the last row is asleep |
| 4 | `SleepStaging.swift:754-756` | post-wake awake-in-bed is emitted only `if block.end > windowEnd` — structurally impossible when they are equal (measured: 0.0 m trailing awake) |
| 5 | `RingSession.swift:1846-1847` | `stagedSegments.map(\.end).max()` becomes the stored in-bed end |

Truncation sweep on 08-09 (dropping trailing real records only):

```
cut 08:30 → last record 08:28:51 → wake 08:30:51 → asleep 537m
cut 09:06 → last record 09:03:51 → wake 09:05:51 → asleep 572m
cut 09:30 → last record 09:28:51 → wake 09:30:51 → asleep 597m
none      → last record 09:56:21 → wake 09:58:21 → asleep 625m
```

`block.end − lastRecord = +120 s` on all eight cuts; trailing awake 0.0 m on all eight; onset never moved.

The `+120 s` itself is **not** a fabrication — `PROTOCOL.md:517` has `[0:4]` as the epoch START (🟢), so a record at 09:56:21 covers 09:56:21–09:58:51 and a block ending 09:58:21 lies inside measured time. Trimming it is worth ≤2 min and would break byte-identity on 8 nights for nothing. **Do not "fix" this by trimming the tail.**

### Why the existing defences don't fire

- **Motion** — the primary `[10:15]` channel is the `[1,1,1,1,1]` placeholder on 271/284 worn epochs, and on **100 % of epochs from 00:03:50 to 09:56:21**. `BulkSleep.motionSource:383` gates the `[15:20]` fallback on `!worn.contains { !$0.motionIsPlaceholder }` — all-or-nothing — and 13 evening epochs veto it.
- **HR** — floor (p12) = 49, `wakeThreshold = 49 + 18 = 67`. Smoothed HR from 09:06 on: `63,62,62,62,60,60,62,62,62,63,62,62,63,64,64,65,65,65,63,60.5,58`. Peaks at 65. Never reaches 67.
- **`markPointOfNoReturnOffset`** — effectively inert: audited over 15 blocks it COMMITS on 1 of 14 distinct blocks and removes 2 minutes. Declined by the terminal-REM vitals guard on 5 nights and by `scanStart <= earliest` on 8 more — in 5 of those `notBefore` equals the **last row index**, making `s > earliest` unsatisfiable, so a single vitals-softened final epoch disables the whole pass by construction.

### 🔴 The vitals guard's premise is REFUTED on this hardware

`SleepStaging.swift:1128-1129` claims the sleep-vitals stream "thins at true wake". Measured **from the true wake**, suffix/body vitals ratio is **1.00 on 08-09** and **0.53 on 08-05** against a 0.50 bar — it declines even a perfectly placed cut on both labelled nights. The doc-comment's supporting number was measured per-HOUR while the code applies the test to the whole suffix; premise and test are on different windows.

Relaxing `hrWakeRescueVitalsFraction` 0.5→1.0 **cannot** fix 08-09 at any setting (0.481 > 0.476) and costs 65 min on 06-27 + 25 min on 06-30.

---

## The lead worth pursuing: the SpO2 duty-cycle cadence

While the ring runs its sleep-measurement program it emits an SpO2 reading every **300 s**, which makes `BulkRecord.layout` alternate sleepVitals/activity 1:1 at 150 s. That program **exits** near wake; off-regime the ring goes to 600 s or off.

⚠️ **Name it correctly.** `raw[8]` is the **SpO2 byte** (PROTOCOL §5.3, APK `spo2` loc 0xb) with `0x12`/`0x13` as "no SpO2 here" sentinels; `layout` is a discriminator *computed* at `BulkSleep.swift:67-88`, not a device mode tag. This is an **SpO2-cadence** rule. Any future change to `layout`'s fall-through silently moves it.

Evidence (🟢, re-derived independently from raw frames by a second decoder, 5650 unique records / 4 rings / 0 byte conflicts):

- sleepV inter-arrival modal at 300 s on every ring: AD 68.3 %, 9F 63.1 %, u3 83.6 %, u4 78.1 %
- same-layout violations on AD: activity-activity 1305 vs sleepV-sleepV **24** — the ring essentially never emits two consecutive SpO2 readings
- violation rate 6.0 % inside the staged sleep window vs 43.6 % outside; split at true wake 3.9 %→23.8 % (08-09), 1.6 %→34.8 % (08-05)
- Mann-Whitney AUC 0.819 / 0.929 (08-09) and 0.937 / 1.000 (08-05) — far above every vitals channel

**Confounds decisively refuted.** 20 sync/drain/BLE-reconnect events sit *strictly inside* violation-free runs, including four successful drains inside one 5.7 h run (06-30 08:55–08:58) and a BLE teardown+reconnect inside 08-09's 191-epoch run. The 3.2× page-boundary enrichment was regime confounding and goes to **0.0 %** under control (147 cross-page pairs inside the quiet regime, zero violations). Battery/charger clean.

Probe result (`lens-nomotion-probe` @ `98f2478`, env-gated, **reference only**):

```
08-09  625m 09:58:21 → 575m 09:08:51   error +52.4 → +2.9 min   (truth 09:06)
08-05  528m 08:52:44 → 485m 08:10:14   error +50.7 → +8.2 min   (truth 08:02)
10 of 13 other blocks byte-identical · Kit suite 1250/0 with the pass active
```

---

## What must be fixed before any of this ships

1. **It oscillates on 08-05 end-to-end.** Measured through the Kit: `473(wit) · 473 · 500(UNW) · 505 · 510 · 478(wit) · 520(UNW) · 485(wit)` — four sign changes, 47-minute swing from 25 minutes of extra data. (The raw signal latches cleanly in isolation; the oscillation appears once the HR confirmation and survival guards are in the loop, so measure **end-to-end**, never the raw estimator.) This is the same defect class that killed the activity-corroboration candidate.
2. **Gap- and jitter-blindness.** The probe's test is `dt == epochSeconds && layout[i] == layout[i-1]`. Dropping one interior epoch makes its neighbours same-layout in **90.5 %** of positions by construction; a gap-blind run was measured bridging 70 missing epochs to manufacture a 25-epoch *daytime* run. 91 of 3432 AD transitions are not exactly 150 s (151–221 s with no record missing), silently suppressing 57 genuine violations. Use `nsteps = round(dt/150)`; when exactly one epoch is missing, infer the absent layout from parity and continue rather than severing or bridging blind. Monte-Carlo dropout on 08-09 at K=12: worst case −57 min at 1 % dropout, −75 at 3 %, −135 at 5 %. A page lost at 08:36 makes the gap-aware rule cut −29.6 min; one lost at 09:06 makes the gap-blind rule cut +17.9 min. The two definitions fail in **complementary** directions.
3. **K=12 sits on a one-epoch margin worth 43 minutes.** K=6/8/10/11 → 09:51 (+45 min); K≥12 → 09:08 (+3 min), because the longest post-wake run on that night is exactly 11 epochs. Five more minutes lying still and the rule would have missed. Round 1's "K=12…40 identical on all 15 blocks" is **false for the raw signal**: 08-04 gives 08:38 at K≤24 and 07:28 at K≥32; 06-29 gives 07:55 at K≤60 and 04:40 at K=80. Either raise K and accept under-correction, or require a second independent trailing-edge witness.
4. **Regime plausibility gate.** 🔴 "the quiet window tracks the sleep window on every ring" is **refuted**. Ring u4 holds the 300 s cadence for a continuous **11.96 h** (timezone-independent — no zone makes that a night; likely continuous SpO2 enabled, rule blind). Ring 9F/Air fragments into 1.3–3.8 h pieces, two per night, and never holds a whole night. Refuse to fire when the cadence covers an implausible share of the day, or never holds for a plausible night. *This also refutes as stated the memory note that FR04 "alternates 0x12/SpO2 epochs 1:1 all night" — the alternation is there, but only in pieces.*
5. **The −75 min on 06-29 is unadjudicated.** The raw estimator removes 75 minutes from an unlabelled night; only the HR no-return confirmation stops it. A guard doing the real work on the night where the signal is most aggressive needs ground truth, not an argument.

### Acceptance criteria

- New `SleepStaging.Tuning` knob where **0 = disabled = byte-identical to master** (the kill-switch shape every other pass in that file uses).
- Structured like `markPointOfNoReturnOffset`: suffix-only by construction, bounded by `onsetSearchEpochs`, gated behind `notBefore`, confirmed by the HR no-return test, reverted unless a consolidated asleep run survives. Ship it as the **locator** the existing pass lacks, not a replacement.
- **Truncation sweep at 5-minute steps on both labelled nights, run end-to-end, must be monotone with zero sign changes.** This is the acceptance test; a candidate that fails it does not ship regardless of its headline number.
- Dropout Monte-Carlo reported alongside it — truncation stability is not dropout stability.
- Corpus table with **correct per-ring timezone** (see gotchas), byte-identical count stated.
- Tests that go red when the call site is deleted (mutation-test them; round 1 caught a mutation-blind suite).
- `swift test` in `ios/OpenCircuitKit` green against the 1250 baseline.

---

## Measured dead ends — do not re-try

| Idea | Measurement |
|---|---|
| **Steps to bound in-bed** | The `[4:6]` step field on `0x10`/`0x87` is a **quarter-hour bucket that resets to 0 at :00/:15/:30/:45** (🟢, three independent captures — `StepAccumulator`'s "current day count" comment is wrong). Read **0 on every one of ~40 frames from 08:41 to 10:04:36** on 08-09 — through the 09:06 wake and past 10:00 while up and using the app. No per-epoch step count exists in the `0x4c` stream. |
| SpO2 as a wake cue | AUC 0.443 / 0.406; SpO2 99 also occurs mid-Deep at 07:26 |
| Respiratory rate | AUC 0.608 / 0.586; a quiet-run rule lands +0.3 min on 08-09, doesn't fire on 08-05, cuts 42 min from 08-04 |
| HRV | AUC 0.789 on one night, 0.502 on the other |
| Confidence byte `[6]` | AUC 0.421 / 0.266 — chance. Its 5-epoch rolling mean is a usable *corroborator* (0.784 / 0.870) but needs a hand-picked absolute cut |
| Sleep-vitals density | **Flips sign** between the two labelled nights (0.546 vs 0.188) |
| Relaxing/re-windowing the terminal-REM vitals guard | Tested at 30 min, 60 min, whole suffix, fraction 1.0 — cannot fix 08-09 at any setting; sheds 65/25/52 min on unlabelled nights |
| `idle`/unworn layout | 2 idle epochs in the whole 08-09 night, both at 22:12–22:14 |
| Skin temp as a wake marker | Real −1.85 °C dip at 09:06, but the same night fires at 05:58 mid-sleep (50 % precision); on 08-04 the big event is 26 min *after* the true wake. Live-only, 0 % coverage in the pre-wake 2 h on 2 of 9 nights |
| `[15:20]` motion tail as a "last quiet run" | Non-zero on ~32 % of epochs all night → no quiet run ever exists → never fires |
| "End of the longest violation-free run" (naive cadence) | Catastrophic: 06-29 → 04:38 (−4 h 28 m), 08-04 → 07:25 (−1 h 29 m). It is "**last** run ≥ K" + the HR confirmation that makes it safe |
| Lowering `wakeHRMarginBPM` | 18→12 changes nothing; 8 collapses the onset. No value works |
| Fixing `constantFiller` alone | Moves 08-09 by **one minute** (09:58 → 09:57) |

---

## Adjacent defects found on the way (each deserves its own issue)

1. **Bedtime is "came off the charger."** The ring was charging 22:23–22:35 on 08-08 — that is the `22:14 ──0.4h──> 22:36` archive hole — and the app's stored bedtime is exactly 22:36. Same class of defect at the leading edge.
2. **A real 32-min in-bed fragment is silently discarded.** 08-09 stages as 2 fragments; the 13-epoch first one is dropped for being under `minSleepDuration` (1 h), and the reported bedtime is then read off the *next* fragment's first record.
3. **The #41 wear gate is inert on the staged path.** `SleepStaging.classifyContiguous:545` calls `mainSleep` with **no** `temperatures:` argument, while the coarse `sleepSegments` path (`RingSession.swift:1680`) does pass them. The asymmetry is a live trap even though it wouldn't have helped this night.
4. **`constantFiller` is dead code.** Fires on **0 of 15** sources; the only run that reaches `[15:20]` gets there via the #184 degenerate branch. The failure it was written for is present on 4 of 8 Gen-2 nights. Worth fixing on its own merits — but see the dead-ends table: it is not the wake fix, and a naive replacement is dangerous (below).
5. **`StepAccumulator` header comment is factually wrong** — see the steps row above.
6. **`layout` fall-through wart.** Any `raw[8]` that is not `0x12`/`0x13` becomes `.sleepVitals`, including 29 records with `[8]` of `0x00`/`0x0a`/`0x0e`/`0x11` — impossible SpO2 values, 1.63 % of non-sentinel epochs. `0x11` (13 occurrences, just under the `0x12` sentinel) looks like an unrecognised third sentinel (🟡). Six manufacture a spurious cadence violation each.
7. **PROTOCOL.md is stale in two places.** `BulkSleep.swift:64`'s `.activity // payload in [15:22]` is the #93 claim that `PROTOCOL.md:371-378` retracted on 2026-06-17; `desktop/decode_bulk.py:33` carries the same stale line. Separately, `[15:23)` decodes as **five 12-bit big-endian magnitudes nibble-packed + a 4-bit flag** (60 + 4 = 8 bytes), which resolves PROTOCOL.md:364's unexplained "V2 map bit-packs acti_counts as length 7.5, info = 0.5 B" (🟢, carry test 0.0032/0.0031/0.0032 against the 1/256 prediction, controls 0.17–0.42, 6804 records). Note `motionIntensityTailIsZero` mis-reports quiet on 13.6 % of the epochs it calls quiet — but three shipped calibrations depend on the legacy predicate, so **do not redefine it**; add a separate accessor.

---

## Branches (all local, nothing merged or pushed)

| Branch | Status |
|---|---|
| `lens-nomotion-probe` @ `98f2478` | The SpO2-cadence probe. **Reference only** — env-gated, instrumented, gap-blind. Start here. |
| `fix/sleep-wake-evidence-kit` @ `6d09093` | ✅ Safe. 18/18 byte-identical, 1275 tests / 0 failures, 10/10 mutations caught. Classifies a wake as `.unwitnessed` when it is the data edge (corpus split is parameter-free: 0 \| ≥2 records after the wake, nothing at 1 — 8 unwitnessed / 8 witnessed). ⚠️ Its original "prerequisite for a wake detector" justification is **refuted** — the cadence pass leaves the in-bed envelope byte-identical, so `sameCoverage` fires and master's merge already accepts the shorter night (verified: 08-09 625→575 and 08-05 528→485 both land on master). On its own it buys −2 min on two nights, plus one real case: with the detector active, master's merge freezes 08-05 at a wrong 520 m while the fixed policy delivers 485 m. |
| `fix/sleep-wake-evidence-persistence` @ `4971c2a` | Stacked on the above. App/UI/schema half — `wakeWitnessed` + SwiftData **V5→V6**. ⚠️ **HOLD** until device-validated: failure mode is `wipeAndRecoverForeground` deleting un-resyncable history on every install. Migration tests pass on the simulator only. |
| `fix/motion-source-dead-primary-run` @ `e7d4ebf` | ❌ **REFUTED, do not ship.** Its `magnitude >= positive[p80] ? 16 : 1` mapping is a *rank* with no absolute floor: on a night whose non-zero tail epochs are mostly quiet noise, every one is declared movement. Measured end-to-end at −158, −212, −265 min and, at a 50 % rate, a night reduced to **zero** (`mainSleep` nil → blank sleep card). Also mis-attributed: on 2 of 3 moved nights the effect came from re-scoping the night so the *old* filler branch fired. |
| `fix/sleep-offset-activity-corroboration` @ `b73c9dd` | ❌ **REFUTED, do not ship.** Oscillates with sync time (three deterministic sign changes, 47-min swing) because `activityStaysAlive` measures density to the end of whatever has drained. Forcing `corroborated = true` gives byte-identical output on all 18 nights — the corroboration has **zero** measured selectivity; it is guard-deletion (a known dead end) plus a 3-epoch anchor on one night. |
| `fix/sleep-unwitnessed-wake` @ `578cf7e` | Superseded by the two `sleep-wake-evidence-*` branches, which fix its blocking no-asleep-bound defect. |

---

## Gotchas for whoever picks this up

- **The corpus is multi-ring and multi-timezone, and the timezone is functional.** 18 nights across **5 distinct rings / 4+ people** (proven byte-for-byte: two files are the same ring iff their shared epoch counters carry identical 23 bytes). Only 10 nights are the AD/Gen2 primary user. The 9F/Air ring is **Europe/Paris**; `SleepWindow.isOvernightBlock` tests the block midpoint against 21:00–09:00 **local** via `Calendar.current`. Replaying the Paris nights in New York produces a false blank sleep card on 08-02-b and a phantom 19:31 "evening night" on 08-07.
- **Ground truth is n=2**, both the same person and the same ring (08-09 → 09:06, 08-05 → 08:02). Below this project's own evidence bar. State it in any PR.
- **Flat-motion fixture trap.** A constant motion value de-floors to STILL, so a synthetic "awake" fixture stages as sleep and assertions go vacuous. Express awake as elevated HR, never as motion. A second trap: a flat 55 bpm fixture sits on the resting floor forever and `sleepVitalsRescue` walks the block to the last record whatever motion says.
- **Replay harness:** `DiagnosticsFrameImport.records(fromDiagnosticsText:)` parses a diagnostics export straight into `[BulkRecord]` → `BulkSleep.latestNightRecords` → `SleepStaging.classify`. Keep harnesses **out of the repo** — they read private health data from `~/Downloads`.

## The one thing that would move this further than any code

**Log the alarm/wake time for a week.** Seven labelled nights would settle the K question, adjudicate the 06-29 −75 min case, and lift the whole effort above n=2. `SleepEditLabel` already turns a user's own sleep edits into supervised labels, and the replay infrastructure now exists.

perezjuanj:
Closes #190.

## The defect

When the primary motion channel is a flat placeholder and the morning HR rise never clears `floor + wakeHRMarginBPM`, every sleeper-side pass misses the wake, the staged night runs to the last record, and the reported wake becomes **`lastRecord + 120 s`** — whenever the user synced. 🟢 Truncating the 2026-08-09 capture in 49 five-minute steps moved reported sleep **367 → 625 min**, one +5 per step, with no byte of physiology changed.

## The fix

`SleepStaging.markCadenceWakeOffset` asks the **ring** instead of the sleeper. While the ring runs its sleep-measurement program it reads SpO2 every 300 s, so 150-second epochs alternate `sleepVitals`/`activity` 1:1; the duty cycle exits at wake. Wake is the epoch after the **last** unbroken alternating run of at least `cadenceWakeQuietEpochs`.

The protocol fact is now in `docs/PROTOCOL.md` §5.3 — 5650 records / 4 rings / 0 byte conflicts, violation rate 6.0 % inside the staged sleep window vs 43.6 % outside, AUC 0.819–1.000, with the sync/drain/BLE/page-boundary confounds re-tested under regime control.

### Both labelled nights

```
08-09  625m 09:58:21 → 575m 09:08:51    error +52.4 → +2.9 min   (truth 09:06)
08-05  528m 08:52:44 → 485m 08:10:14    error +50.7 → +8.2 min   (truth 08:02)
```

## Answering the five blockers in the issue

**1. It oscillated on 08-05 end-to-end.** Root-caused, and it was never the locator. The locator returned `end=232@08:07:44 → cut 08:10:14` on *every* truncation cut; what flickered was **`lastVitalsSoftened`** in `notBefore` (nil → 241 → 245 → nil → 245 → nil), which is derived from `tailStart`, a night-wide HR floor, and a rolling median truncated at the data edge. Gating on it produced `478 · 473 · 473 · 497 · 502 · 507 · 512 · 517 · 485` — three sign changes, 47-minute swing. The pass is therefore gated on the second-bout rescue only. The moving-but-asleep restless morning that the softening exists for is still protected, by the HR no-return confirmation (a sleeper at the floor is never cut) — and the 24 existing tests in that family stay green.

**2. Gap- and jitter-blindness.** `round(dt/150)`, one missing epoch bridged by **parity**, anything longer is `.unknown` and can never terminate a trusted run. Dropout Monte-Carlo, 100 trials each:

| night | 1 % | 3 % | 5 % |
|---|---|---|---|
| 08-09 | **100/100** wake identical | 95/100 (4× +2.5 min, 1× declines to master) | 88/100 |
| 08-05 | 96/100 | 94/100 | 83/100 |

Every failure mode is ≤ 10 min or a clean decline back to master. The gap-blind K=12 rule this replaces was −57 / −75 / −135 min at 1/3/5 %.

**3. K sat on a one-epoch margin.** Sweeping K per night gives the closed interval that yields the same cut: 08-09 [12,191] · 08-05 [7,119] · 08-04 [6,27] · 08-02 [2,40] · 06-27 [4,161] · 06-30 [2,137]. **Every night agrees for K ∈ [12, 27]**, bounded below by 08-09's 11-epoch post-wake run and above by 08-04's 27-epoch final run. **K = 20** maximises the worst-case headroom (7 epochs ≈ 17 min either side). A test pins the default inside the measured band.

**4. Regime plausibility gate.** Three declines: no run ≥ K anywhere; a run that reaches the **data edge** (which is what makes a ring left in continuous SpO2 mode a no-op — and is what stops the pass re-manufacturing the artefact it removes); and `cadenceWakeMaxQuietEpochs` (240 ≈ 10 h) for the u4-class 11.96 h regime. 🔴 The last one does not fire anywhere in the corpus (longest trusted run 191 epochs ≈ 7.96 h) and is synthetic-tested only.

**5. The −75 min on 06-29.** It never happens: the **HR no-return confirmation declines it**, and 06-29 comes out byte-identical. Same for 08-07.

## Acceptance

- **Kill switch:** `cadenceWakeQuietEpochs = 0` is **byte-identical to master, 18/18**, verified against a regenerated `CorpusBaseline` (md5 `6663c44ec8234c25a56c364357868d18`, reproduced before use).
- **Truncation sweep, 5-min steps, end-to-end.** 08-09: **0 sign changes**, and the wake locks at 09:08:51 for all 11 cuts past the true wake while master keeps adding +5 min per step. 08-05: the wake is identical on every cut from 08:12 on. ⚠️ 08-05's *asleep total* still carries one −5 min step at 08:17:44 — that is master's night-wide HR floor (p12 over rows that include the post-wake tail) drifting as data arrives, not the locator; master shows the same class of step at 05:32:44 with the pass **off**. Called out rather than papered over.
- **Corpus, per-ring, correct timezones:** 11 of 18 byte-identical. Changed: 08-09 −50, 08-05@13-03 −43, 08-05 export −17, 08-04@12-58 −17, 08-04 export −17, 06-27 −5, 08-02 −2. **The in-bed envelope and the onset are unchanged on every night** — this is a pure trailing-edge pass.
- **Tests:** 1278 / 0 failures (1250 baseline + 28 new). **All 12 mutations of the new code are caught**, including deleting the call site, taking the longest run, dropping the data-edge guard, exact-`dt` equality, removing parity bridging, and reverting the default.
- iOS app target builds.

## ⚠️ What this ships on

**Ground truth is n = 2, the same person on the same ring.** That is below this project's own evidence bar and is why `cadenceWakeQuietEpochs` is a one-line kill switch. The corpus (18 nights / 5 rings / 4+ people) constrains *stability* and *byte-identity*, not accuracy. Logging alarm/wake for a week — `SleepEditLabel` already turns sleep edits into supervised labels — is still the thing that would move this furthest.

Not device-validated yet.

## Not in scope (each deserves its own issue)

The issue's "adjacent defects" list is untouched: bedtime is still "came off the charger", the 32-min in-bed fragment is still discarded, the #41 wear gate is still inert on the staged path, `constantFiller` is still dead code, `StepAccumulator`'s header comment is still wrong, and the `layout` fall-through wart is unchanged. Master's trailing-edge band drift (the −5 min above) is a new one found on the way.

### DIFF
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -428,6 +428,38 @@ time (no 12 h offset in this capture; bears on §5.6/§6.6). Reassemble + decode
   sleep-vitals = "not the idle template AND `[8]` ∉ {`0x12`,`0x13`}", so a sub-87 % desaturation
   still keeps its HR/HRV/SpO2 (the old value-gate dropped the whole epoch — see `BulkSleep.swift`).
 
+**The SpO2 DUTY CYCLE — the ring's own "I am measuring sleep" witness** 🟢 (2026-08-09, #190).
+While the ring runs its sleep-measurement program it takes an SpO2 reading every **300 s**. Epochs
+are 150 s, so consecutive `0x4c` records ALTERNATE sleep-vitals / activity **1:1**, and the
+alternation stops when the program exits — which is at, or just after, final wake. Measured over
+5650 unique records / 4 rings, re-derived independently from raw frames by a second decoder with
+**0 byte conflicts**:
+
+| observation | value |
+|---|---|
+| sleep-vitals inter-arrival, modal bin | **300 s** on every ring (AD 68.3 %, 9F 63.1 %, u3 83.6 %, u4 78.1 %) |
+| consecutive same-template pairs, ring AD | activity-activity 1305 vs sleepVitals-sleepVitals **24** — the ring essentially never emits two SpO2 reads in a row |
+| same-template ("violation") rate | **6.0 %** inside the staged sleep window vs **43.6 %** outside |
+| split at the user-reported wake | 3.9 % → 23.8 % (08-09); 1.6 % → 34.8 % (08-05) |
+| Mann-Whitney AUC vs wake | **0.819–1.000**, far above SpO2 (0.44/0.41), RR (0.61/0.59), HRV (0.79/0.50) or the confidence byte (0.42/0.27) |
+
+Confounds are REFUTED, not assumed: 20 sync, drain and BLE-reconnect events sit strictly INSIDE
+violation-free runs (including four successful drains inside one 5.7 h run, and a BLE
+teardown+reconnect inside a 191-epoch run), and the apparent 3.2× page-boundary enrichment goes to
+**0.0 %** under regime control (147 cross-page pairs inside the quiet regime, zero violations).
+
+⚠️ **Name it correctly.** This is a rule about **`[8]`, the SpO2 byte** — `BulkRecord.layout` is a
+discriminator *computed* from it (`0x12`/`0x13` ⇒ activity), not a device mode tag. Any change to
+that discriminator's fall-through silently moves this signal.
+
+⚠️ **Two caveats before building on it.** (1) It is **jitter- and gap-sensitive**: 91 of 3432 steps
+on ring AD are 151–221 s with no record missing, so an exact `dt == 150` test suppresses 57 genuine
+violations; and dropping ONE interior epoch makes its neighbours same-template in 90.5 % of
+positions by construction. Use `round(dt/150)` and bridge a single hole by parity. (2) It does **not**
+hold for a whole night on every ring: `9F`/Air fragments into 1.3–3.8 h pieces (two per night), and
+`u4` holds the cadence for a continuous **11.96 h**, which no timezone makes a night — most likely
+continuous SpO2 enabled. Consumed by `SleepStaging.cadenceSteps` / `markCadenceWakeOffset`.
+
 **Sleep-vitals fields — confirmed against the RingConn app's readout for the
 2026-06-13 night** (avg HR 68 / HRV 65 ms / SpO2 98 %, low 93 % ~02:30–03:00):
 - **`[4]` = heart rate (bpm)** 🟢. Sleep-window mean ~60, dips to 56–57 in deep-sleep
--- a/ios/OpenCircuitKit/Sources/OpenCircuitKit/Analytics/SleepStaging.swift
+++ b/ios/OpenCircuitKit/Sources/OpenCircuitKit/Analytics/SleepStaging.swift
@@ -302,6 +302,73 @@ public enum SleepStaging {
         /// ~5 ≈ 12.5 min, covering the device-observed 2026-07-18 11-minute dropout.
         public var preOnsetBedtimeMaxGapEpochs: Int
 
+        // --- SpO2-CADENCE wake OFFSET (the "reported wake is when you synced" fix, #190) ---------
+        // Every pass above locates final wake from the SLEEPER (HR, motion, HRV density). When the
+        // primary motion channel is a flat placeholder and the morning HR rise is small, all of them
+        // miss, and the staged night then runs to the LAST RECORD — so the reported wake is
+        // `lastRecord + 120 s`, i.e. whenever the user last synced. 🟢 MEASURED on 2026-08-09
+        // (FR02.018, build 39): truncating the same capture in 49 five-minute steps moved reported
+        // sleep 367 → 625 min, one +5 per step, with no byte of physiology changed.
+        //
+        // This pass locates the wake from the RING instead. While the ring runs its sleep-measurement
+        // program it takes an SpO2 reading every 300 s, which makes consecutive 150-second epochs
+        // alternate sleep-vitals / activity 1:1 (`BulkRecord.layout`). That duty cycle EXITS near
+        // wake. 🟢 CONFIRMED over 5650 unique records / 4 rings, re-derived independently from raw
+        // frames by a second decoder with 0 byte conflicts:
+        //   • sleep-vitals inter-arrival is modal at 300 s on every ring (63–84 %);
+        //   • on the primary ring, activity-activity same-template pairs number 1305 against just 24
+        //     sleepVitals-sleepVitals — the ring essentially never emits two SpO2 reads in a row;
+        //   • the same-template ("violation") rate is 6.0 % inside the staged sleep window and 43.6 %
+        //     outside it; split at the user-reported wake it goes 3.9 %→23.8 % (08-09) and
+        //     1.6 %→34.8 % (08-05);
+        //   • Mann-Whitney AUC 0.819–1.000 on the two labelled nights, far above every vitals channel
+        //     (SpO2 0.44/0.41, RR 0.61/0.59, HRV 0.79/0.50, confidence byte 0.42/0.27).
+        // Sync/drain/BLE confounds are REFUTED: 20 sync, drain and reconnect events sit strictly
+        // INSIDE violation-free runs, including four successful drains inside one 5.7 h run and a BLE
+        // teardown+reconnect inside 08-09's 191-epoch run. Cross-page enrichment goes to 0.0 % under
+        // regime control (147 cross-page pairs inside the quiet regime, zero violations).
+        //
+        // ⚠️ NAME IT CORRECTLY. `raw[8]` is the SpO2 byte (PROTOCOL §5.3; APK `spo2` loc 0xb) with
+        // `0x12`/`0x13` as "no SpO2 here" sentinels. `layout` is a discriminator COMPUTED at
+        // `BulkSleep.layout`, not a device mode tag — this is an SpO2-CADENCE rule, and any change to
+        // that discriminator's fall-through silently moves it.
+
+        /// Minimum length, in epochs, of the unbroken sleep-vitals/activity ALTERNATION that the wake
+        /// locator will trust as "the ring was running its sleep-measurement program here". The wake
+        /// candidate is the epoch immediately after the LAST such run. **0 DISABLES the pass entirely
+        /// — byte-identical to the pre-cadence staging** (the regression escape hatch every other
+        /// pass in this file carries: `motionAwakeVitalsHalfWindow`, `hrWakeRescueCeilingBPM`,
+        /// `preOnsetBedtimeReachEpochs`, `offsetNoReturnSpreadFraction`).
+        ///
+        /// 🟢 The admissible band is MEASURED, not guessed. Sweeping K over the 16-night local corpus
+        /// gives, per night, the closed interval of K that yields the SAME committed cut:
+        /// 08-09 [12,191] · 08-05 [7,119] · 08-05(export) [6,119] · 08-04 [6,27] · 08-04(export)
+        /// [6,27] · 08-02 [2,40] · 06-27 [4,161] · 06-30 [2,137]. The binding constraints are
+        /// 08-09's 11-epoch post-wake quiet run (K must exceed it) and 08-04's 27-epoch final run
+        /// (K must not exceed it), so every night in the corpus agrees for K ∈ [12, 27]. 20 is the
+        /// value that MAXIMISES the worst-case headroom over that corpus (7 epochs ≈ 17 min either
+        /// side); 12 — the value a naive `minConsolidatedSleepEpochs` reuse would pick — sits on a
+        /// ONE-epoch margin worth 43 minutes on 08-09.
+        ///
+        /// 🟡 The corpus behind that band is 16 nights / 5 rings, but the two nights carrying WAKE
+        /// GROUND TRUTH are the same person on the same ring (08-09 → 09:06, 08-05 → 08:02). That is
+        /// below this project's own evidence bar; see `docs/RUNBOOK_SLEEP_GROUNDTRUTH.md` and
+        /// `SleepEditLabel` for the supervised labels that should re-fit it.
+        public var cadenceWakeQuietEpochs: Int
+
+        /// Upper bound, in epochs, on a quiet run the locator will trust. A run LONGER than this is
+        /// not a night's sleep-measurement program — it is a ring left in continuous SpO2 mode, and
+        /// its "end" carries no information about when anyone woke up.
+        ///
+        /// 🔴 This gate does NOT fire anywhere in the local corpus (the longest trusted run is 191
+        /// epochs ≈ 7.96 h, on 08-09). It exists because the corpus DID produce the failure it guards
+        /// against: ring `u4` holds the 300 s cadence for a continuous 11.96 h — timezone-independent,
+        /// no zone makes that a night. On that ring the run happens to reach the data edge, which the
+        /// locator already declines; this bound is what stops the same ring being cut at a stray
+        /// violation once more data drains in behind it. 240 ≈ 10 h, i.e. above every real night
+        /// measured here and below the u4 regime.
+        public var cadenceWakeMaxQuietEpochs: Int
+
         public init(awakeMotion: Int = 15,
                     deepHRPercentile: Double = 0.42,
                     remHRPercentile: Double = 0.86,
@@ -332,7 +399,9 @@ public enum SleepStaging {
                     minConsolidatedSleepEpochs: Int = 16,
                     deepBaselineMarginBPM: Double = 18,
                     preOnsetBedtimeReachEpochs: Int = 24,
-                    preOnsetBedtimeMaxGapEpochs: Int = 5) {
+                    preOnsetBedtimeMaxGapEpochs: Int = 5,
+                    cadenceWakeQuietEpochs: Int = 20,
+                    cadenceWakeMaxQuietEpochs: Int = 240) {
             self.awakeMotion = awakeMotion
             self.deepHRPercentile = deepHRPercentile
             self.remHRPercentile = remHRPercentile
@@ -364,6 +433,8 @@ public enum SleepStaging {
             self.deepBaselineMarginBPM = deepBaselineMarginBPM
             self.preOnsetBedtimeReachEpochs = preOnsetBedtimeReachEpochs
             self.preOnsetBedtimeMaxGapEpochs = preOnsetBedtimeMaxGapEpochs
+            self.cadenceWakeQuietEpochs = cadenceWakeQuietEpochs
+            self.cadenceWakeMaxQuietEpochs = cadenceWakeMaxQuietEpochs
         }
 
         public static let `default` = Tuning()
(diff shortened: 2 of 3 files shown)


perezjuanj:
⚠️ **Correction to this PR's body, measured after the fact — see #197.**

This PR attributed the residual −5 min step on the 08-05 truncation sweep to "master's night-wide HR floor (p12 over rows that include the post-wake tail) drifting as data arrives". **That is refuted.** The HR floor is 53 bpm at *both* 08:12:44 (478 min) and 08:17:44 (473 min) — it does drift 53 → 54, but one cut later, and the total does not move when it does. #194 independently measured that pinning the floor across 514 truncation cuts on 6 nights changes nothing at all.

The real mover is `BulkSleep.motionIntensityFallbackMagnitudes`'s **p80 rank over the drained record set**: its cut oscillates 249 → 247 → 247 → 249 across those cuts, and interior epochs 89 and 193 sit exactly on it (magnitude 1 = still at 249, magnitude 16 > `awakeMotion` 15 = awake at 247). Reported asleep tracks it exactly: 478 · 473 · 473 · 478.

Nothing in the merged code changes — the cadence locator and both pinned wake times are unaffected. Only the stated *cause* of the residual was wrong. Tracked in #197, which also refutes a second candidate (`tailStart`) proposed while chasing it.

Issue #204, opened 2026-08-11
The night summary can fail to persist silently while naps succeed — the Sleep card then shows live staging that no stored night backs


perezjuanj:
## Summary

A Gen 3 tester's schema-3 Data Export (build 39, range 2026-08-09 → 2026-08-12) contains:

- `"sleep": []` — empty
- no `sleepSessions` key at all
- `naps` — **3 entries**, all on 08-10
- 3337 samples, 1762 daytimeTemperatures, 324 stepSamples, 24 historySyncEvidence

…while the Sleep card was displaying a full night for 08-10→11 (8h31m asleep, a hypnogram, stage averages).

**No `StoredSleepSummary` row exists for that night.** The card is rendering live in-memory staging.

## Proof it is write-side, not an export defect

`provenance` in the file carries exactly the 7 base keys — `samples`, `stepSamples`, `daytimeTemperatures`, `sleep`, `daily`, `naps`, `historySyncEvidence` — and **none** of `sleepSessions`, `sleepSessions.summary`, `sleepSessions.osa`, `sleepSessions.hypnogram`, `sleepSessions.coverage`. Those keys are added only when the session array is non-empty (`ExportEngine.swift:861`, `:891-900`), so the array was empty **at build time**, not lost in serialization.

Ruled out:

- **Export cap** — the only cap is `ExportBuilder.maxExportDays = 365`, a no-op on a 3-day range, and it drops the *oldest days*, never a whole table. `ExportEngine` has no row cap or `prefix()`.
- **Range/predicate mismatch** — sleep filters on the night bucket (`LocalStore.swift:1478-1484`). An 08-10→11 night keys to 08-10 or 08-11 under either keying scheme; both are strictly interior to `[08-09, 08-12)`. No timezone shift of a few hours reaches a boundary.
- **The night classified as a nap** — the 3 naps are 14:37, 16:59 and 17:20 local, all midday, none crossing the night.

## The mechanism

`RingSession.persistSleepAndSteps` (~`:1838-1876`) defers the night while writing naps unconditionally. Its own comment describes this exact symptom:

> ⚠️ The deferral covers the NIGHT only. `persistNaps` below runs off THIS drain's buffer …

The `saveSleepSummary` call is `try?`, so the throw is swallowed silently. Three write-side paths produce "no night, naps fine":

[...]

Issue #195, opened 2026-08-09
Decoder + docs: `layout` fall-through and the 0x11 sentinel, dead `constantFiller`, the [15:23) nibble-packed decode, and two stale PROTOCOL.md claims


perezjuanj:
Split out of #190. Four small, related items in the `0x4c` decoder and its documentation. ⚠️ **Read the interaction warning at the bottom first** — item 1 can silently move a signal that shipped in #191.

## 1. `layout`'s fall-through, and a probable third sentinel

`BulkRecord.layout` returns `.sleepVitals` for **any** `raw[8]` that is not `0x12`/`0x13` and is not the idle template. That is deliberate (#39: a genuine desaturation below 87 % must not be thrown away) but it currently swallows **29 records** whose `[8]` is `0x00`/`0x0a`/`0x0e`/`0x11` — impossible SpO2 values, 1.63 % of non-sentinel epochs.

`0x11` in particular (13 occurrences, sitting just under the `0x12` sentinel) looks like an **unrecognised third "no SpO2 here" sentinel** — 🟡, needs settling from the wire, not from plausibility.

Consequence today: six of those records manufacture a spurious cadence violation each.

## 2. `constantFiller` is dead code

`BulkSleep.swift:383`'s `let constantFiller = !worn.contains { !$0.motionIsPlaceholder }` is all-or-nothing, and it fires on **0 of 15** sources in the corpus. The failure it was written for is present on **4 of 8** Gen-2 nights — the only run that ever reaches `[15:20]` gets there via the #184 degenerate branch instead.

⚠️ **A naive replacement is dangerous and this is measured.** Branch `fix/motion-source-dead-primary-run` was REFUTED: its `magnitude >= positive[p80] ? 16 : 1` mapping is a *rank* with no absolute floor, so on a night whose non-zero tail epochs are mostly quiet noise every one is declared movement — end-to-end −158, −212, −265 min, and at a 50 % rate a night reduced to **zero** (`mainSleep` nil → blank sleep card). On 2 of 3 moved nights the effect was mis-attributed: it came from re-scoping the night so the *old* filler branch fired.

Also note #190 measured that fixing `constantFiller` alone moves the 08-09 night by **one minute**. It is worth fixing on its own merits. It is not a sleep-duration fix, and it must not be sold as one.

## 3. `[15:23)` decodes as five 12-bit nibble-packed magnitudes 🟢

[...]

PR #208, opened 2026-08-18
fix(sleep): desk-activity wake gate — stop a sedentary evening being …

jameswhickmott:
…staged as sleep

The primary [10:15] motion channel does not register desk work: it reads the SAME value at a keyboard as it does in deep sleep. 🟢 MEASURED on two ground-truthed evenings (FR02.018, Gen 2, Australia/Melbourne) in which the wearer was sitting at a computer — 0 of 61 epochs above the motion baseline on 2026-08-17, and 0 of 157 during that night's real sleep.

So the detector cannot break the run. `mainSleepBlock` welds evening to night, and the HR gate — which judges a whole period on its MEDIAN — cannot trim a HEAD: the 2026-08-17 night came out as ONE unbroken 14h03m "sleep" block, 17:49:54 → 07:53:12, with sleep credited from 19:40 against a wearer who was at a computer until 00:30. No existing knob fixes it. Sweeping the coarse HR margin 25 → 10 does not move the block start at all and at 8 the whole night vanishes; sweeping the staged wake margin 18 → 6 moves onset by 27 min while shedding 3h24m of real sleep. It is not a tuning problem.

The ring DID record the movement, in `activityMagnitudes` ([15:23), 🟢 #195). Share of epochs reading ALL-ZERO, both nights:

    real sleep                                72 % / 79 %
    at a computer   [ground truth]             0 % /  2 %
    up and about                               0 % / 10 %
    evening dozing  [ground truth]            14 %

So threshold that share over a centred 45-min rolling window and force the epoch ACTIVE below 0.35. Unlike the wear and HR gates this acts PER EPOCH and BEFORE segmentation, so the run splits — which is what a sedentary evening welded onto the night needs. One-directional, like both of them: it can only ever turn sleep into active.

Replayed byte-exactly against both archives:

    2026-08-18   onset 19:40 → 01:20   asleep 10h09m → 6h02m   (truth: computer until 00:30)
    2026-08-16   onset 01:00 → 01:02   asleep  8h26m → 8h24m   (already correct; stays correct)

Sweeping the threshold 0.20 → 0.60 moves onset by only ~22 min on either night, so the knob is not hypersensitive — deliberately unlike the two margins above.

⚠️ This deliberately relaxes the "consumers may use it only through `BulkSleep.motionSource`" constraint on the intensity tail. That constraint exists to stop the two motion channels being MIXED into one magnitude scale — the measured trap that moved 15 of 18 corpus nights by −100…+45 min, and moved the same night in opposite directions in two bundles. This gate never computes a magnitude: it thresholds the SHARE of a binary predicate whose floor is the number zero. A record set cannot move where zero is.

⚠️ 🟡 THE THRESHOLD IS FIT ON TWO NIGHTS, ONE WEARER. 0.35 sits between the highest awake reading measured (0.14) and the lowest sleep reading (0.72). Wide, but two nights is two nights — to be re-fit against accumulated `SleepEditLabel` corrections. Evening dozing is genuinely intermediate and will be trimmed back to its stillest core.

Kill switch: `ActivityPeriod.deskWakeZeroShareThreshold = 0` disables the gate outright and is byte-identical to pre-#204, mirroring every other pass in SleepDetection. All 1464 pre-existing tests pass unchanged — synthetic fixtures carry zero activity magnitudes, so the gate is inert on them, and that is pinned by a test rather than left to luck.

Also adds `SleepReplay`, a CLT triage tool (mirroring RingKitVerify) that replays an export's raw records through the shipping pipeline and prints the decision trace — how this was diagnosed, and how it will be validated. Requires `sleepHRFloor` to be public so the tool reports the real floor rather than a copy that could drift.

### DIFF
--- a/ios/OpenCircuitKit/Package.swift
+++ b/ios/OpenCircuitKit/Package.swift
@@ -23,5 +23,9 @@ let package = Package(
         // CLT-friendly verifier: `swift run RingKitVerify` works without Xcode,
         // asserting the same real-capture fixtures. Stopgap until Xcode is present.
         .executableTarget(name: "RingKitVerify", dependencies: ["OpenCircuitKit"]),
+        // Field-triage tool: replays a wearer's exported 0x4c records through the shipping
+        // sleep pipeline and prints the decision trace. Read-only; changes no behaviour.
+        // `swift run SleepReplay <export.json>` — see Sources/SleepReplay/main.swift.
+        .executableTarget(name: "SleepReplay", dependencies: ["OpenCircuitKit"]),
     ]
 )
--- a/ios/OpenCircuitKit/Sources/OpenCircuitKit/Analytics/SleepDetection.swift
+++ b/ios/OpenCircuitKit/Sources/OpenCircuitKit/Analytics/SleepDetection.swift
@@ -58,6 +58,22 @@ public struct HeartRateSample: Sendable, Equatable {
     }
 }
 
+/// One epoch's verdict from the ring's OWN activity magnitudes — `BulkRecord.activityMagnitudes`,
+/// the five 12-bit nibble-packed values in `[15:23)` (🟢 layout MEASURED, PROTOCOL.md §5.3 / #195).
+/// `quiet == true` means all five read zero: the ring itself saying nothing moved this epoch.
+///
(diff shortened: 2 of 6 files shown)


jameswhickmott:
Sorry didn't mean to create this PR haven't managed to test it

Issue #197, opened 2026-08-09
Staged sleep is a function of sync timing on any night using the [15:20] fallback: `motionIntensityFallbackMagnitudes` maps through a p80 RANK over the whole record set


perezjuanj:
Found while verifying #194, and independently corroborated from the opposite direction in #195. **This supersedes two earlier attributions of the same symptom, both of which were wrong — mine in #191 and the one recorded in #194.**

## The defect

`BulkSleep.motionIntensityFallbackMagnitudes` maps the `[15:20]` intensity tail to a movement magnitude through

```swift
let activeCut = degenerate
    ? otsuIntensityCut(positive)
    : positive[Int((Double(positive.count - 1) * 0.80).rounded())]   // ← a RANK over the whole set
```

`positive` is every positive tail sum **in the record set being staged**. So the "is this movement?" threshold is a **quantile of however much history has drained**, with no absolute floor. Add records, the rank moves, and every epoch in the night is remapped.

## Measured, on 2026-08-05 (AD/Gen2 FR02.018), end-to-end through the shipped pipeline

Truncating the capture in 5-minute steps, staging each cut:

| cut | `positive.count` | `p80cut` | epoch 89 / 193 magnitude | interior awake | **asleep** |
|---|---|---|---|---|---|
| 08:12:44 | 87 | **249** | 1 (still) | 3 | **478** |
| 08:17:44 | 89 | **247** | 16 (> `awakeMotion` 15 → awake) | 5 | **473** |
| 08:22:44 | 90 | 247 | 16 | 5 | 473 |
| 08:27:44 | 91 | **249** | 1 | 3 | **478** |

Two interior epochs sit exactly on the moving cut and flip awake↔asleep with it. The reported night moves **±5 min for no physiological reason at all** — only because of when the user synced. Same defect class as #190, different mechanism.

At 08:52:44 `motionSource` flips `tail(false)` → `primary` outright and the night settles at 485.

## The two attributions this replaces

- **#191's PR body said**: "master's night-wide HR floor (p12 over rows that include the post-wake tail) drifting as data arrives". 🔴 **Refuted** — the floor is 53 bpm at *both* 08:12:44 (478 min) and 08:17:44 (473 min). It does drift (53 → 54), just one cut later, and the total does not move when it does.
[...]

Issue #184, opened 2026-08-02
Gen 2 Air (FR04): sleep never stages — the 0x4c motion channel has a third idle floor the still-detector can't read


perezjuanj:
## Symptom

Tester on build 34 with a **RingConn Gen 2 Air (FR04.009)**: *"not tracking my sleep."* Zero stored sleep summaries across two nights, blank sleep card, and no `cursor sleep` line in either diagnostics bundle (that key is only written once a night reaches Health).

## ⚠️ Scoping caveat — this is the first FR04 device we have ever decoded

Every byte layout, threshold and fixture in this repo was validated on **FR02.018 (Gen 2)** and **FR05.008 (Gen 3)** only. This ring trips `hasFirmwareMismatch` because it is not the pinned FR02.018 (`FirmwareInfo.swift:47`).

**There is currently no way to scope a fix to FR04.** 🟢 The generation is cosmetic: `RingGeneration` is declared once (`FirmwareInfo.swift:24`) and read in exactly three production places — `DeviceInfoView.swift:53`, `DiagnosticsReport.swift:39`, `HistoryFrameCapture.swift:107` — all string formatting. A grep for every comparison form returns **zero** production hits. `FirmwareInfo.swift:12` states it outright: *"no decode path branches on generation."* The Kit never receives the generation; `SleepStaging.Tuning` has ~30 knobs and no device field, and the detection layer has no tuning type at all (non-injectable `static let`s at `SleepDetection.swift:78-87`).

**So any change to `motionStillThreshold` or the de-flooring changes Gen 2 and Gen 3 behaviour too.** Building a per-model gate is part of the work, not a given.

**Precedent argues against a generation switch.** The one time a generation branch was genuinely needed — Gen 3's elevated/drifting motion floor (Gen 2 ≈1, Gen 3 ≈15→24→39) — it was solved with a self-calibrating rolling local floor (`SleepDetection.swift:204-253`), explicitly a no-op for Gen 2. Prefer adapting to the signal over keying on the label.

## Not the usual suspects

🟢 Sync, decode and storage are all healthy:

- Archive complete — 655 and 606 epochs, **no hole >6 min anywhere in either night**.
- Decode fine — HR valid on 358/360 records, physiologically plausible.
- Storage fine — cursors load and advance to the latest epoch.
- `restageFromArchive()` re-stages from the persisted archive on **every** connect (`RingSession.swift:990`), so this is not a missed drain window.

[...]

PR #211, opened 2026-09-02
fix(sleep): stage the night on a raised primary-motion floor (Gen 2 Air, FR04.011)

ericjypark:
> **Revised 2026-09-02** after the reporter supplied a real FR04.011 archive and I replayed it. The branch was landing the right *rescue* on the wrong *premise*, and both conjuncts of the gate turned out to be measuring the wrong thing — including one that rejected the very archive this PR was written for. See [**What changed in this revision**](#what-changed-in-this-revision). The first revision's analysis is left in the history (`e8dcd56`) rather than squashed away, since the correction is the interesting part.

## What

Adds a **third rejection reason** to `BulkSleep.motionSource` — a *raised, wandering* primary-motion floor — and a third `MotionSource` case that reads the decoded `[15:23)` `activityMagnitudes` instead of the byte-aligned `[15:20]` tail.

## Why: a Gen 2 Air on FR04.011 stages nothing at all

Reported from a Gen 2 Air on firmware **FR04.011**, build 52. Every history drain ends `nightRowOutcome = noStagedSegments` with `stagedSleepSegments = 0` — 19 of 24 evidence rows over two nights; the remaining rows have no night row at all. HR/HRV/RR/SpO₂ are recorded fine, with sleep-vitals epochs present all night. It is a blank card on a complete archive, the #184 symptom with a different cause.

**Field evidence** (numbers only; no raw epochs are included in this PR):

- Over **715 worn epochs**, the primary `[10:15]` byte takes the `01` "still" baseline value on only **356 of 3575** sub-samples. The rest cluster around **17, 41–53, 71, 84–86** *during obviously motionless sleep*.
- Per-night medians **≈ 75–95**, with the per-night 25th percentile at **54** on one night and **1** on the other.
- 🟢 **Within an epoch the five sub-samples are nearly *equal*** — median intra-epoch spread **1 count**, p90 **6**, over the ring's own motionless epochs. (They are *exactly* equal only ~8 % of the time, which is what the first revision measured and then over-read; see below.)
- 🟢 **The wander is *between* epochs**: the pedestal's level moves **1 → 126** across the night, and the per-epoch minimum runs p10 **1** / p50 **40** / p90 **93**.
- The decoded 12-bit `activityMagnitudes` channel is clean on the *same* records: the per-epoch magnitude sum is **exactly 0 for the median night epoch** (p50 = 0, p90 ≈ 700–800), rises to **100–450** at postural turns and **> 1000** at the final wake, and its hourly profile matches the HR/HRV dynamics.

## How the detector fails today

`motionSource` has two reasons to reject the primary channel, and FR04.011 satisfies neither:

- **Constant filler** — needs every worn epoch to be `motionIsPlaceholder`. The level moves, so no.
- **Non-expressive (#184, FR04.009)** — needs both `degenerateMaxQuietStillFraction` *and* `degenerateMinSlotOrderFraction`. The second gate **correctly fails**: no slot pair keeps a fixed ordering, so `slotOrderConsistency` rightly reports "this is not an instrumentation template". It isn't one.

So the run keeps `.primary`. `motionAboveLocalFloor` subtracts a per-window level, which cancels a flat plateau **whose level is stable across the floor's ~30-minute window** — Gen-2 `01`, Gen-3 `0f`, the drifting `16→24→39`. FR04.011's pedestal is flat *inside* an epoch but wanders further and faster than that window can track, so the p10 floor sits well below the current level and a large residual survives de-flooring. After de-flooring only **32 %** of quiet epochs read still, against the **≥ 70 %** (`gravityStillFraction`) that `detect()` requires. The phantom movement swamps detection and no block is staged.

## The new rule

`primaryFloorIsRaised(worn)` — same method as `primaryMotionIsDegenerate` (condition on the ring's own "nothing moved" verdict, then ask whether the primary agrees), with two deliberate differences:

1. The verdict is **`activityMagnitudesAreZero`**, the layout-correct decode from #195, not the byte-aligned `[15:20]` window. This branch *reads* the decoded magnitudes, so it is gated on the same numbers it will consume.
2. The "the residual is instrumentation, not movement" proof is a **raised, wandering floor** rather than a fixed template.

Gates, in order:

| gate | value | why |
|---|---|---|
| enough ring-verified motionless time to judge | `degenerateMinQuietEpochs` (24 epochs = 1 h), **reused** | below that there is no stageable night anyway — and those epochs *are* the magnitude channel saying "nothing moved", so this doubles as the "is the substitute any good?" test |
| primary refuses to read still **after de-flooring** | `degenerateMaxQuietStillFraction` (0.50), **reused from #184**, applied to `primaryChannelIsStillAfterFloor` | this is what keeps every flat or drifting placeholder night on the primary path — such a channel de-floors to zero and reads still |
| the floor is off baseline | median per-epoch `min(raw[10..<15])` ≥ **16** | clears the highest documented healthy placeholder (Gen-3 `0f` = 15) by one count |

The load-bearing addition is **`primaryChannelIsStillAfterFloor(_:)`**: it pushes the raw primary channel through the *production* `ActivityPeriod.motionAboveLocalFloor` and asks, per epoch, whether ≥ `gravityStillFraction` of the five residuals fall under `motionStillThreshold`. It measures what `detect()` actually consumes, and reuses the detector's own two constants rather than restating them, so the predicate and the detector it predicts cannot drift apart. It reads `raw[10..<15]` directly rather than via `motionTimeline`, which would re-enter `motionSource` and recurse.

The branch is evaluated **last**, after both tail reasons decline, and carries the same "the channel must actually carry movement" conjunct against the channel it selects — so **no input that reached a verdict before this change changes verdict**.

Selected runs map `Σ activityMagnitudes` onto the same `0 / 1 / 16` alphabet the tail fallback already emits, so detection, staging and the movement chart consume one scale whichever channel a run picked.

## What changed in this revision

Replaying the reporter's real archive (715 worn epochs, one night plus the preceding day) broke both conjuncts of the gate `e8dcd56` shipped:

**(a) The stillness conjunct measured the wrong axis.** `e8dcd56` gated on `motionResolvesStillness`, which asks whether an epoch's five sub-samples sit within `motionStillThreshold` of *one another*, reasoning that a flat plateau cancels against the rolling floor at any level. That reasoning holds only while the plateau's **level** is stable across the floor's own window — and here it is not. Because the pedestal *is* flat inside an epoch, the proxy fires on **78 %** of quiet epochs, scores the unusable channel "still", and the gate **rejected the real FR04.011 archive** while happily accepting the synthetic fixture. The proxy answers a question about one epoch; the failure lives in the drift between epochs. Replaced with the de-floored measurement above.

**(b) The share-of-worn conjunct was scope-dependent, and is deleted.** `raisedFloorMinZeroMagnitudeFraction ≥ 0.40` was measured on **night** epochs but evaluated wherever `motionSource` is called — and `latestNightRecords` hands it the whole **30 h `EpochArchive` union**, ~22 h of which is an awake day. On the real archive it reads **0.278** over the union against **0.567** over the night alone, so the gate opened or stayed shut depending on how much daytime happened to have drained: an answer that depends on *when you synced*. An absolute quorum is scope-stable; a share of the window is not. The surviving `degenerateMinQuietEpochs` already covers the duty it was carrying.

**Net effect on new constants: this PR now adds one 🟡 constant, not two.** `primaryChannelIsStillAfterFloor` introduces none — it borrows `gravityStillFraction` and `motionStillThreshold` from the detector it predicts.

Replaying the reporter's archive with the corrected gate stages **47 segments, onset 01:17, wake 08:40, efficiency 0.90**.

### Honest limits

- `raisedFloorMinMedianQuietMinimum = 16` is a **conservative floor, not a measured separation point**, tagged 🟡 in source per `CLAUDE.md`. It rests on one device.
- ⚠️ **The margin on the surviving stillness conjunct is thin.** On this single capture it clears `degenerateMaxQuietStillFraction` by only **0.028** at night-only scope. That is real but slim, and I would rather flag it than tune it against the one archive I have.
- 🙏 **A second FR04.011 capture is genuinely wanted** — ideally a different wearer, or at least a different night. It would do more for confidence in this branch than any further adjustment of the two thresholds. If anyone reading has a Gen 2 Air on FR04.011, a raw epoch archive would be very welcome.
- `activityMagnitudeActiveCut = 700` is an **observed p90, not a fitted threshold**, and the doc comment says so. #197's finding stands untouched: a real seam needs `[15:23)` magnitudes *and* supervised labels. The `0`-versus-positive split does the load-bearing work regardless — a still epoch maps to `0` and a stir to `1`, both below `motionStillThreshold`.

I deliberately did **not** widen `primaryMotionIsDegenerate` or touch the three calibrations fitted to the byte-aligned population (`degenerateMaxQuietStillFraction`, `degenerateMinSlotOrderFraction`, `hrvPoolingNoiseFloorMs`).

## Test coverage

`FR04RaisedPrimaryFloorTests`, now **12 tests**. All records are **synthetic**, built byte-by-byte in the test — including the nibble packing, which is round-trip asserted against `activityMagnitudes` so the magnitude assertions cannot go vacuous. No captured health data.

New in this revision, a **`wanderingPedestalNight` fixture built to the field statistics** — flat within an epoch (spread ≈ 1), level stepping between the observed plateau clusters:

- `testWanderingPedestalFixtureMatchesTheFieldStatistics` — pins the property that broke the first revision: on this fixture the intra-epoch proxy *does* read "still" above `degenerateMaxQuietStillFraction`, so a regression to `motionResolvesStillness` fails here loudly.
- `testWanderingPedestalSelectsTheMagnitudeChannelAtEveryScope` — asserts the verdict at **both** scopes production evaluates `motionSource` at (the night slice **and** the archive union). This is the regression (b) would otherwise reintroduce.
- `testWanderingPedestalNightStages` — the night stages, with a > 5 h main block.

The original fixture is **kept, not replaced**: it draws all five sub-samples independently, so it exercises a genuinely different shape, and both must select the magnitude channel. Retained alongside: a classic `01`-baseline night still selects `.primary` and still stages; both gates in isolation; and the sub-hour guard.

### Results

`swift test` in `ios/OpenCircuitKit`: **1838 tests, 11 skipped, 1 failure** — and that one failure is pre-existing:

```
SleepStagingTests.testMidNightWASOIsImmuneToTheRescueButTheMorningTailIsNot
XCTAssertEqualWithAccuracy failed: ("211.5") is not equal to ("182.5") +/- ("5.0")
```

I re-verified it reproduces **identically (211.5 vs 182.5) with my source changes stashed**, so it is not caused by this branch, and I have not touched it. Everything else passes, including the corpus replays, `SleepMonotonicityTests`, and every existing Gen-2/Gen-3 and FR04.009 motion fixture.

<details>
<summary>⚠️ Heads-up: the test target does not compile on Swift 6.2.1 (unrelated to this PR)</summary>

On `swift-driver 1.127.14.1 / Apple Swift 6.2.1 (swiftlang-6.2.1.4.8)`, arm64-apple-macosx26.0, a clean `master` fails to **build the test target** with type-checker timeouts in three test files, before any test runs:

- `SleepMonotonicityTests.swift:172` — the chained `+` string literal in the `XCTAssertEqual(collapses, [], …)` message
- `SleepConfidenceCopyTests.swift:136` — `silentFor: 4 * 3600` needing a `TimeInterval` annotation
- `AutomaticWorkoutDetectionTests.swift:28` — the `sample(cursor:bpm:steps:)` map closure

I patched these **locally only** to be able to run the suite, then reverted them — **none of it is in this branch**, since it is unrelated to the bug and you may prefer a different fix or a pinned toolchain. Happy to send it as a separate PR if useful. (This is also why the test numbers above come from a locally-patched checkout.)
</details>

## Docs

`docs/PROTOCOL.md` §5.3 tabulates all three known shapes of `[10:15]` against the channel the detector reads on each. **This revision corrects the raised-floor row and its paragraph**, which had the first revision's error baked in: it described the variation as within-epoch ("counts wander freely… the sub-samples differ epoch to epoch") and claimed the channel's minimum "never returns to `01`". Both are wrong — the variation is *between* epochs, and the per-epoch minimum runs p10 1 / p50 40 / p90 93. What defeats de-flooring is the *rate* the level drifts at, not a floor permanently stuck off `01`. Numbers only, no raw epochs.

I did **not** add a release-notes entry: the repo writes `docs/RELEASE_NOTES_bNN.md` per build at release time, and there is no `b52` file yet — I would rather not pre-empt how you write it. I also did not bump the build number.

### DIFF
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -597,15 +597,24 @@ run's own distribution, never from a device or firmware string:
 |---|---|---|---|
 | baseline | `01` (Gen 2), `0f`=15 (Gen 3), or a drifting `16→24→39` plateau — **constant within the epoch**, so a rolling local floor cancels it at any level | Gen 2 / Gen 3 | `[10:15]` (unchanged) |
 | fixed template | a two-level intra-epoch **step** (slots 0–1 ≈ 27.6, slots 2–4 ≈ 34.9 on *every* epoch) ± 2 noise; survives de-flooring because the step is *inside* one epoch | Gen 2 Air, **FR04.009** | `[15:20]` intensity tail |
-| **raised floor** | counts wander **freely** in 40–90 with the occasional `1`; the sub-samples differ epoch to epoch and no slot ordering is phase-locked, so it is *not* a template — but the channel's own **minimum never returns to `01`** | Gen 2 Air, **FR04.011** | `[15:23)` decoded magnitudes |
+| **raised floor** | a pedestal that is **flat *within* an epoch** (median intra-epoch spread **1 count** on motionless epochs) but whose **level wanders between epochs**, 1 → 126 across the night — further and faster than the ~30-min rolling local floor can track, so de-flooring does *not* cancel it; no slot ordering is phase-locked, so it is not a template either | Gen 2 Air, **FR04.011** | `[15:23)` decoded magnitudes |
 
 The **raised-floor** case is new. 🟡 Measured on **one device, two nights, 715 worn epochs**: the
 primary byte takes value `1` on only 356 of 3575 sub-samples and otherwise clusters around 17,
 41–53, 71 and 84–86 *during obviously motionless sleep*; per-night medians ≈ 75–95 with the
-per-night 25th percentile at 54 on one night and 1 on the other. All five sub-samples are equal on
-only ≈ 8 % of epochs, so neither `motionIsPlaceholder` nor `motionResolvesStillness` fires and the
-FR04.009 gate (`slotOrderConsistency`) correctly refuses it — the variation really is free, it just
-never comes back down. On device the symptom is a total staging failure: every history drain ends
+per-night 25th percentile at 54 on one night and 1 on the other.
+
+🟢 **The variation is BETWEEN epochs, not within one** — an earlier revision of this note had that
+backwards. All five sub-samples are *exactly* equal on only ≈ 8 % of epochs, but they are *nearly*
+equal on nearly all of them: median intra-epoch spread **1 count**, p90 **6**, over the ring's own
+motionless epochs. So `motionResolvesStillness`, which asks whether the five sit within
+`motionStillThreshold` of one another, **does** fire — on **78 %** of quiet epochs — and it is
+wrong to do so, because the channel it is blessing cannot arbitrate stillness. Nor does the minimum
+"never return to `01`": the per-epoch minimum runs p10 **1** / p50 **40** / p90 **93**. What breaks
+de-flooring is the *rate* the pedestal's level drifts at, not a floor stuck permanently off `01` —
+after `motionAboveLocalFloor` only **32 %** of quiet epochs read still (the detector needs ≥ 70 %).
+`motionIsPlaceholder` and the FR04.009 gate (`slotOrderConsistency`) do correctly refuse it. On
+device the symptom is a total staging failure: every history drain ends
 `nightRowOutcome = noStagedSegments` with 0 staged segments while HR/HRV/RR/SpO₂ record all night.
 
 🟡 The **decoded** `[15:23)` magnitudes are clean on the same records: the per-epoch magnitude sum
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -589,6 +589,32 @@ baseline (still/asleep) and spike at arousals/turns — the per-epoch **stillnes
 Phase 5 `SleepDetection` needs (likely the IMU stream; no separate `0x47` accel needed).
 Baseline `01` = "still", not "unworn".
 
+🟡 **…on the firmware that HAS a baseline. Three shapes of `[10:15]` are now known, and two of
+them cannot express stillness at all** — `BulkSleep.motionSource` picks the run's channel from the
+run's own distribution, never from a device or firmware string:
+
+| shape | `[10:15]` while motionless | seen on | detector reads |
+|---|---|---|---|
+| baseline | `01` (Gen 2), `0f`=15 (Gen 3), or a drifting `16→24→39` plateau — **constant within the epoch**, so a rolling local floor cancels it at any level | Gen 2 / Gen 3 | `[10:15]` (unchanged) |
+| fixed template | a two-level intra-epoch **step** (slots 0–1 ≈ 27.6, slots 2–4 ≈ 34.9 on *every* epoch) ± 2 noise; survives de-flooring because the step is *inside* one epoch | Gen 2 Air, **FR04.009** | `[15:20]` intensity tail |
+| **raised floor** | counts wander **freely** in 40–90 with the occasional `1`; the sub-samples differ epoch to epoch and no slot ordering is phase-locked, so it is *not* a template — but the channel's own **minimum never returns to `01`** | Gen 2 Air, **FR04.011** | `[15:23)` decoded magnitudes |
+
+The **raised-floor** case is new. 🟡 Measured on **one device, two nights, 715 worn epochs**: the
+primary byte takes value `1` on only 356 of 3575 sub-samples and otherwise clusters around 17,
+41–53, 71 and 84–86 *during obviously motionless sleep*; per-night medians ≈ 75–95 with the
+per-night 25th percentile at 54 on one night and 1 on the other. All five sub-samples are equal on
+only ≈ 8 % of epochs, so neither `motionIsPlaceholder` nor `motionResolvesStillness` fires and the
+FR04.009 gate (`slotOrderConsistency`) correctly refuses it — the variation really is free, it just
+never comes back down. On device the symptom is a total staging failure: every history drain ends
+`nightRowOutcome = noStagedSegments` with 0 staged segments while HR/HRV/RR/SpO₂ record all night.
+
+🟡 The **decoded** `[15:23)` magnitudes are clean on the same records: the per-epoch magnitude sum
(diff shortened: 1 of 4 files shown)


ericjypark:
**Revised after replaying a real night** — please re-read the description, the analysis changed.

The reporter sent a raw FR04.011 archive, and replaying it broke the premise the first revision rested on. I had the pedestal's variation *inside* the epoch; it is actually *between* epochs. Within an epoch the five sub-samples are nearly equal (median spread 1 count), so the `motionResolvesStillness` conjunct fired on 78 % of quiet epochs and **the gate rejected the very archive this PR was written for**. And the `raisedFloorMinZeroMagnitudeFraction ≥ 0.40` conjunct was measured on night epochs but evaluated on the whole 30 h archive union (0.278 union vs 0.567 night), so its verdict depended on how much daytime had drained.

`c4fcb08` fixes both: a new `primaryChannelIsStillAfterFloor` measures stillness *through* the production `motionAboveLocalFloor` — i.e. what `detect()` actually consumes — reusing `gravityStillFraction` and `motionStillThreshold` rather than restating them, and the scope-dependent share conjunct is deleted in favour of the absolute quorum already there. Net: the PR now adds **one** 🟡 constant instead of two. Replay stages 47 segments, onset 01:17, wake 08:40, efficiency 0.90.

Left as a follow-up commit rather than a squash, since the wrong turn seemed worth keeping in the record.

Two things worth your scepticism: the surviving stillness conjunct clears its threshold by only **0.028** at night-only scope, and all of this still rests on **one device**. A second FR04.011 capture would be worth more than any further tuning — happy to wait on that before you merge if you would rather.

